PR #10, opened 2017-07-13
[WIP] CLOUD-1735 support additional maven servers

bdecoste:


os-eap-s2i/added/s2i/assemble
```diff
@@ -21,6 +21,31 @@ if [ -f $LOCAL_SOURCE_DIR/configuration/settings.xml ]; then
...
+  if [ "${#repo_ids[@]}" -ne "0" ]; then
+    local count=0
+    for repo_id in ${repo_ids[@]}; do
+      if [ -n "${repo_usernames[$count]}" ] && [ -n "${repo_passwords[$count]}" ]; then
+        local xml="\n\
+          <server>\n\
+            <id>${repo_id}</id>\n\
+            <username>${repo_usernames[$count]}</username>\n\
+            <password>${repo_passwords[$count]}</password>\n\
+          </server>\n\
```
knrc: This would result in the usernames/password being stored in the image which is likely not what is expected, it may be better just to refer to the environment variables within the settings.xml

tremes:
I guess this one could be closed since the related issue has been closed as "Won't do".

PR #14, opened 2017-07-13
CLOUD-1529 SCRIPT_DEBUG for dynamic resources

bdecoste:


dynamic-resources/dynamic_resources.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+if [ "${SCRIPT_DEBUG}" = "true" ] ; then
```
bdecoste: Is that a problem? I think that's a benefit.

rcernich: Let me rephrase, what are the side effects, if any, of this being incorporated into the eap (and layered products) launch scripts?  This changes seems innocuous, but it affects more than just our config scripts.

PR #15, opened 2017-07-13
CLOUD-794 add Connector server config

bdecoste:


rcernich:
The issue seems to be about adding server name, but the commit includes a bunch of extra stuff for configuring ssl keys.  Is the ssl configuration for another issue?

bdecoste:
prepareEnv() was missing

os-jws-launch/added/launch/http.sh
```diff
@@ -0,0 +1,17 @@
...
+  unset JWS_SERVER_NAME
+}
+
+function configure() {
+  configure_http
+}
+
+function configure_http() {
+  if [ -n "$JWS_SERVER_NAME" ]; then 
+    sed -i "s|redirectPort=\"8443\"|redirectPort=\"8443\" server=\"${JWS_SERVER_NAME}\"|" $JWS_HOME/conf/server.xml
```
maschmid: I would be much happier if we'd define the JWS_SERVER_NAME operation as "if JWS_SERVER_NAME is specified, it adds the 'server' attribute to all HTTP connectors that didn't already have the attribute set"  and then use a proper XML transformation for it... the current behavior generally assumes no customizations have been done on the server.xml...

PR #17, opened 2017-07-13
CLOUD-1593 add QUEUE_MEMORY_LIMIT

bdecoste:


os-amq-launch/added/configure.sh
```diff
@@ -182,6 +182,12 @@ function configureIoExceptionHandler() {
     fi
 }
 
+function configurePolicy() {
+  if [ -n "$AMQ_QUEUE_MEMORY_LIMIT" ]; then
+    sed -i "s|memoryLimit=\"1mb\"|memoryLimit=\"${AMQ_QUEUE_MEMORY_LIMIT}\"|" "$CONFIG_FILE"
```
rcernich: This should be more robust, e.g. instead of \"1mb\", maybe \"[^\"]\"

os-amq-launch/added/configure.sh
```diff
@@ -182,6 +182,12 @@ function configureIoExceptionHandler() {
     fi
 }
 
+function configurePolicy() {
+  if [ -n "$AMQ_QUEUE_MEMORY_LIMIT" ]; then
+    sed -i 's|memoryLimit="[^"]*"|memoryLimit="'"${AMQ_QUEUE_MEMORY_LIMIT}"'"|' "$CONFIG_FILE"
```
rcernich: Does this tie us to the formatting in the config file, e.g. single vs double quotes?

bdecoste: That's replacing memoryLimit="anything" with memoyLimit="$AMQ_QUEUE_MEMORY_LIMIT"

rcernich: right, i just want to make sure it works even if the text is different, e.g. ``memoryLimit = '256mb'``, notice the spaces and single quotes vs. double quotes.  reference: https://www.w3.org/TR/REC-xml/#NT-Attribute

I suspect we can safely ignore newlines around the ``=``, but i think we need to accommodate single quotes and spaces., e.g. ``memoryLimit *= *("[^"]*")|('[^']*')`` which may require enabling extended regex (-r flag)

bdecoste: I don't think we need to worry about either single quotes or spaces around the = since the default activemq.xml never does either. I've never seen any of our xml do either. But I will change it to your regex.

PR #18, opened 2017-07-13
CLOUD-1762 threaddump on liveness failure

bdecoste:


rcernich:
These changes look fine, but since we already have a DEBUG_SCRIPT flag for the liveness probe, maybe we should use that instead of introducing a new variable.  WDYT?

bdecoste:
That's fine with me. I'll make the change.

rimolive:
I'm ok too.

PR #22, opened 2017-07-18
[WIP] CLOUD-1862 support generic security-domains

bdecoste:


os-datavirt/added/standalone-openshift.xml
```diff
@@ -525,14 +525,14 @@
             <preparedplan-cache infinispan-container="teiid-cache"/>
             <transport name="embedded"/>
             <transport name="odata">
-                <authentication security-domain="teiid-security"/>
+                <authentication security-domain="##ODATA_SECURITY_DOMAIN##"/>
```
rcernich: Is this something we'd want to make generic, e.g. TRANSPORTS=ODATA, ODATA_TRANSPORT_NAME=odata, ODATA_TRANSPORT_SECURITY_DOMAIN=teiid-security.  Obviously, we're limited as to what we might do for existing transports, so maybe another option is to modify teiid-security to handle this?

os-eap-launch/added/launch/security-domains.sh
```diff
@@ -5,17 +5,77 @@ function prepareEnv() {
...
+  unset ${prefix}_LOGIN_MODULE_CODE
+  unset ${prefix}_LOGIN_MODULE_MODULE
+
+  for option in $(compgen -v | grep -s "${prefix}_MODULE_OPTION_"); do
+    unset ${option}
+  done
 }
 
 function configure() {
+  configure_legacy_security_domains
```
rcernich: Another option is to define mappings in backward-compatibility.sh using variable expression trickery.  Just an idea.

os-eap-launch/added/launch/security-domains.sh
```diff
@@ -5,17 +5,77 @@ function prepareEnv() {
...
+      fi
+      security_domain="$security_domain >"
+
+      local options=$(compgen -v | grep -s "${domain_prefix}_MODULE_OPTION_")
+     
+      for option in $(echo $options); do
+        option_name=$(echo "${option}" | sed -e "s/${domain_prefix}_MODULE_OPTION_//g")
+        option_value=$(find_env $option)
+
+        if [ "$option_name" == "keycloak_config_file" ]; then
```
rcernich: I see your problem now.  This is not a problem for datasources and resources because those are generally setter names and you can't have dashes or dots in method names. hmm....

os-eap-launch/added/launch/security-domains.sh
rcernich: We need to find a better solution.  Hard-coding this stuff doesn't align with a generic solution.

os-datavirt/added/standalone-openshift.xml
bdecoste: I don't think so - Teiid exposes JDBC, ODBC, REST/ODATA, and "embedded". We are covering and exposing all but the last.

os-eap-launch/added/launch/security-domains.sh
bdecoste: Hardcoding what? The option names that contain dots or dashes? I'm looking at alternatives to using properties files and ENVs for passing in name/values.

os-eap-launch/added/launch/security-domains.sh
bdecoste: Right. Didn't run into this in any testing for datasources or resource adapters, but we hit it with both SSO and LDAP module-options.

os-datavirt/added/standalone-openshift.xml
bdecoste: I guess we could make it generic to cover any possible future transports, but there won't be additional ones for the current version of Teiid

rcernich: I was speaking to it being generic and jdv is a first class consumer, defining transports for odata, etc.

os-eap-launch/added/launch/security-domains.sh
rcernich: I think this was supposed to show up under the next comment, e.g. keycloak_config_file.

rcernich:
We need to find a better way to generically define options.  Hard-coding special names is no better than simply having a custom set of env's for each module type.

os-eap-launch/added/launch/security-domains.sh
rimolive: The biggest issue we have here is the lack of standardization of all Security Domains, where some properties are camel case, others are dotted and others are dashed. Unless we use some kind of markdown-style ( e.g. __ for dash substitution and ___ for dot substitution), I don't see another way to make this piece of code generic.

os-datavirt/added/standalone-openshift.xml
rimolive: @bdecoste, is it required having all transports configured? What if the customer want to use only one of them?

os-eap-launch/added/launch/security-domains.sh
rcernich: I don't think there's a good option for escaping these.  We could try something like _OPTION_NAME_1 and _OPTION_VALUE_1, where the suffixes for name/value pairs is consistent (1, 2, foo, whatever).  The other option is to create separate vars for SSO and LDAP.  The latter is probably the most user friendly, since users don't need to go looking up the option names and there's less risk of suffix mismatches.

jmtd:
superceded by #67

PR #25, opened 2017-07-19
CLOUD-1880 DataGrid should use common getfiles script

bdecoste:


bdecoste:
Yes, https://github.com/jboss-openshift/cct_module/pull/39 contains the changes to getfiles that fixes the broken json logging. But that PR doesn't change datagrid.

PR #31, opened 2017-07-20
CLOUD-1841 eliminate empty access.log file

bdecoste:


rcernich:
Doesn't this work apply to EAP 6.4 too?

spolti:
@rcernich as I am working on a related jira, I'll use the same pattern that @bdecoste used.

PR #35, opened 2017-07-21
CLOUD-1840 allow users to define jms destination jndi bindings so can…

bdecoste:
… be accessed remotely


spolti:
This PR looks a little bit old, can we close this?

PR #37, opened 2017-07-24
CLOUD-1848 CLOUD-1849 Add security-realm to management configuration …

bdecoste:
…when admin_user/password is specified


os-eap70-openshift/added/standalone-openshift.xml
```diff
@@ -70,7 +70,7 @@
             </logger>
         </audit-log>
         <management-interfaces>
-            <http-interface http-upgrade-enabled="true" console-enabled="false"><!-- ##MGMT_IFACE_REALM## -->
+            <http-interface http-upgrade-enabled="true" console-enabled="false" ##MGMT_IFACE_REALM##>
```
rcernich: Is there a better way to do this?  We have other issues popping up about our version of standalone.xml not being valid xml.

os-eap7-launch/added/launch/admin.sh
```diff
@@ -15,7 +15,12 @@ function configureEnv() {
 }
 
 function configure_administration() {
-  if [ -n "${ADMIN_PASSWORD}" ]; then
-        $JBOSS_HOME/bin/add-user.sh -u "$ADMIN_USERNAME" -p "$ADMIN_PASSWORD"
+  if [ -n "${ADMIN_PASSWORD}" -a -n "$ADMIN_PASSWORD" ]; then
```
maschmid: should one of these be a check for ADMIN_USERNAME?

os-eap-probes/added/probes/probe/dmr.py
```diff
@@ -49,6 +49,8 @@ def __readConfig(self):
         self.port = 9990 + int(os.getenv('PORT_OFFSET', 0))
         self.user = os.getenv('ADMIN_USERNAME')
         self.password = os.getenv('ADMIN_PASSWORD')
+        if self.password != "" and self.user == "":
```
maschmid: this only works if ADMIN_USERNAME is set to an empty string, if it is not set at all, it will not be set to "eapadmin", as it should..

bdecoste: If you set just ADMIN_PASSWORD=whatever in the template then you get as ENVs in the pod:
 
sh-4.2$ env | grep ADMIN
ADMIN_USERNAME=
ADMIN_PASSWORD=whatever

The probe now detects that condition and sets ADMIN_USERNAME=eapadmin. It detects ADMIN_USERNAME as an empty string, not as null

tests/features/eap/eap_common.feature
```diff
@@ -9,23 +9,23 @@ Feature: Openshift EAP common tests (EAP and EAP derived images)
...
        | ADMIN_USERNAME          | admin2            |
        | ADMIN_PASSWORD          | lollerskates11$   |
-       | JAVA_OPTS_APPEND            | -Dfoo=bar         |
+       | JAVA_OPTS_APPEND        | -Dfoo=bar         |
     Then container log should contain WFLYSRV0025
-     And run /opt/eap/bin/jboss-cli.sh -c --error-on-interact --no-local-auth --user=admin2 --password=lollerskates11$ deployment-info in container once
+     And run /opt/eap/bin/jboss-cli.sh -c --error-on-interact --no-local-auth --user=admin2 --password=lollerskates11$ deployment-info in container and immediately check its output contains activemq-rar
      # We expect this command to fail, so make sure the return code is zero, we're interested only in output here
-     And run /opt/eap/bin/jboss-cli.sh -c --error-on-interact --no-local-auth deployment-info || true in container and immediately check its output does not contain activemq-rar
+     And run sh -c '/opt/eap/bin/jboss-cli.sh -c --error-on-interact --no-local-auth deployment-inf || true' in container and immediately check its output contains Unable to authenticate
```
maschmid: typo? "deployment-inf", instead of "deployment-info"

PR #38, opened 2017-07-24
CLOUD-1817 better logging for multiple jars

bdecoste:


os-java-run/added/run-java.sh
```diff
@@ -41,10 +46,11 @@ auto_detect_jar_file() {
       ls *.jar | grep -v '^original-'
       exit 0
     fi
+
+    echo >&2 "ERROR: Neither \$JAVA_MAIN_CLASS nor \$JAVA_APP_JAR is set and ${nr_jars} found in ${dir} (1 expected)"
```
Ladicek: When rendered, the error looks like

```
ERROR: Neither $JAVA_MAIN_CLASS nor $JAVA_APP_JAR is set and 2 found in /deployments (1 expected)
```

I think clearly the word "JARs" is missing. Can we change to "${nr_jars} JARs found in ${dir}"?

Ladicek:
Thanks :-)

os-java-run/added/run-java.sh
```diff
@@ -3,6 +3,11 @@
 # Fail on a single failed command
 set -eo pipefail
 
+if [ "${SCRIPT_DEBUG}" = "true" ] ; then
```
rcernich: We need to make sure this is documented in the image too, if it isn't already (i.e. image.yaml).

PR #41, opened 2017-07-26
cloud-1955 fix rest ssl with new realm

bdecoste:


os-jdg7-launch/added/launch/infinispan-config.sh
```diff
@@ -590,18 +590,18 @@ function configure_infinispan_endpoint() {
           fi
 
           if [ -n "${HTTPS_NAME}" -a -n "${HTTPS_PASSWORD}" -a -n "${HTTPS_KEYSTORE_DIR}" -a -n "${HTTPS_KEYSTORE}" ] ; then
-            encryption="<encryption $rest_security_realm />"
+            encryption="<encryption security-realm=\"ApplicationRealm\" />"
```
rcernich: Does this need to match the value in SECDOMAIN_NAME used in authentication-config.sh?

bdecoste: SECDOMAIN_NAME only applies when authentication is enabled. The ssl config is the same in either case, but I can change to use the newly generated realm IFF authentication is enabled.

PR #48, opened 2017-08-02
CLOUD-1962: Update AMQ tests for 6.3 R4.

jmtd:
Bump the version string for the first test.

Remove the detailed version string check for the second test; the point
of that check is as a "guard" for the steps that take place afterwards
(to make sure that we wait until start up has reached a certain point
before we try probe tests); therefore, we don't need the full version
in the check.


jmtd:
These changes correspond to the (not merged) changes in this PR for the parent image https://github.com/jboss-container-images/jboss-amq-6-image/pull/7

PR #55, opened 2017-08-07
[CLOUD-1965] [SSO7] Support clean shutdown on TERM

kyguy:

### DIFF
--- a/os-sso/added/openshift-launch.sh
+++ b/os-sso/added/openshift-launch.sh
@@ -70,8 +70,18 @@ JAVA_PROXY_ARGS="$(proxy_options)"
 
 echo "Running $JBOSS_IMAGE_NAME image, version $JBOSS_IMAGE_VERSION"
 
+# TERM signal handler
+function clean_shutdown() {
+  echo "*** JBossAS wrapper process ($$) received TERM signal ***"
+  $JBOSS_HOME/bin/jboss-cli.sh -c ":shutdown(timeout=60)"
+  wait $!
+}
+
+trap "clean_shutdown" TERM
+
 if [ -n "$SSO_IMPORT_FILE" ] && [ -f $SSO_IMPORT_FILE ]; then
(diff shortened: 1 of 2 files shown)


os-sso/added/openshift-launch.sh
```diff
@@ -70,8 +70,18 @@ JAVA_PROXY_ARGS="$(proxy_options)"
...
 echo "Running $JBOSS_IMAGE_NAME image, version $JBOSS_IMAGE_VERSION"
 
+# TERM signal handler
+function clean_shutdown() {
+  echo "*** JBossAS wrapper process ($$) received TERM signal ***"
+  $JBOSS_HOME/bin/jboss-cli.sh -c ":shutdown(timeout=60)"
+  wait $!
+}
+
+trap "clean_shutdown" TERM
```
rcernich: The changes for EAP had a variable to disable this.  Why the difference between the two?

kyguy: The CLI_GRACEFUL_SHUTDOWN variable prevents EAP from being shutdown by a TERM signal, so that JBossAS can be terminated cleanly by the shutdown script(/opt/eap/bin/jboss-cli.sh -c ":shutdown(timeout=60)"). Since SSO7 runs EAP7, it appears the CLI_GRACEFUL_SHUTDOWN script was probably intended to be added here too. 

When I was making the EAP changes , I worked around CLI_GRACEFUL_SHUTDOWN script so that it would continue to function properly. If the purpose of CLI_GRACEFUL_SHUTDOWN is to prevent JBossAS shutdown on TERM signals, then CLI_GRACEFUL_SHUTDOWN should be added to the SSO launch script. But If the sole purpose of the CLI_GRACEFUL_SHUTDOWN script is to cleanly shutdown JBossAS(as the name suggests) then it can be safety removed from the EAP launch script.  I was leaning toward the latter but I wasn't completely sure, so I left the functionality intact

rcernich: Got it.  Thanks!

PR #56, opened 2017-08-07
[CLOUD-1784] - Make the accessLog valve configurable

spolti:


spolti:
Fixed little typos on the comments on the access_log_valve.sh

PR #58, opened 2017-08-08
cloud-1185 remove java tmp perf data dir owned by 185 after s2i

bdecoste:


tests/features/openshift_s2i.feature
```diff
@@ -81,3 +81,8 @@ Feature: Openshift S2I tests
     Then XML file /home/jboss/.m2/settings.xml should have 1 elements on XPath //ns:mirror[ns:id='mirror1'][ns:url='http://127.0.0.1:8080/repository/internal/'][ns:mirrorOf='external:*']
     And XML file /home/jboss/.m2/settings.xml should have 1 elements on XPath //ns:mirror[ns:id='mirror_bar'][ns:url='http://127.0.0.1:9090/repository/other/'][ns:mirrorOf='*']
     And s2i build log should contain WARNING: Variable "willfail_MAVEN_MIRROR_URL" not set. Skipping maven mirror setup for the prefix "willfail".
+
+  Scenario: Check java perf dir owned by jboss
+    Given s2i build https://github.com/jboss-openshift/openshift-examples from spring-eap6-quickstart
+      | variable             | value                                      |
+    Then run ls -l /tmp in container and immediately check its output for jboss root
```
rcernich: Would it be better to actually run jstat instead of checking the dir perms?  I think this tests the actual problem reported in the jira.

bdecoste: I tried jstat in the test and it always fails. I spent a bunch of time trying to figure out why and eventually gave up. It runs correctly in the pod, but not in the test for some reason. I'll revisit.

PR #61, opened 2017-08-11
[OpenJDK] Correct permissions of home directory

kyguy:

### DIFF
--- a/os-java-run/install_as_root
+++ b/os-java-run/install_as_root
@@ -21,4 +21,5 @@ chmod 755 /opt/run-java/run-java.sh \
     /opt/run-java/debug-options  \
     /opt/run-java/container-limits \
     /opt/run-java/java-default-options \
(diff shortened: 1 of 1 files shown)


os-java-run/install_as_root
```diff
@@ -21,4 +21,5 @@ chmod 755 /opt/run-java/run-java.sh \
     /opt/run-java/debug-options  \
     /opt/run-java/container-limits \
     /opt/run-java/java-default-options \
-    /opt/run-java/proxy-options
+    /opt/run-java/proxy-options \
+    /home/jboss
```
rcernich: @goldmann, do you mind reviewing this?  This is going to be related to the passwd.sh script, which only seems to be included with the jboss product images and not the openjdk image.  Given that, I wonder if this should be included some other place, maybe even the base jboss java image.  WDYT?

PR #65, opened 2017-08-16
[CLOUD-1524] - Recognize JAVA_CORE_LIMIT

spolti:
Recognize JAVA_CORE_LIMIT env var setting or replace it with CONTAINER_CORE_LIMIT in the documentation

### DIFF
--- a/tests/features/eap/eap_gc.feature
+++ b/tests/features/eap/eap_gc.feature
@@ -85,8 +85,45 @@ Feature: Openshift OpenJDK GC tests
   # CLOUD-459 (override default heap size)
   Scenario: CLOUD-459 Check for adjusted default heap size
     When container is started with args
-      | arg       | value                        |
-      | mem_limit | 1073741824                   |
+      | arg       | value                         |
+      | mem_limit | 1073741824                    |
       | env_json  | {"INITIAL_HEAP_PERCENT": 0.5} |
     Then container log should match regex ^ *JAVA_OPTS: *.* -Xms256m\s
       And container log should match regex ^ *JAVA_OPTS: *.* -Xmx512m\s
+
+  Scenario: CLOUD-1524, test JAVA_CORE_LIMIT
+    When container is started with env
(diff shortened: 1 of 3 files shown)


maschmid:
(copying my comment from https://issues.jboss.org/browse/CLOUD-1524)

I think this is wrong..

CONTAINER_CORE_LIMIT should never be set explicitly by the user (it is set by the dynamic-resources/dynamic_resources.sh script if CPU resource limits are set)
JAVA_CORE_LIMIT should be the env var to set if one want to override whatever the CONTAINER_CORE_LIMIT turns out to be...

tests/features/eap/eap_gc.feature
```diff
@@ -90,3 +90,28 @@ Feature: Openshift OpenJDK GC tests
...
+    Then container log should match regex ^ *JAVA_OPTS: *.* -XX:ParallelGCThreads=3\s
+     And container log should match regex ^ *JAVA_OPTS: *.* -Djava.util.concurrent.ForkJoinPool.common.parallelism=3\s
+     And container log should match regex ^ *JAVA_OPTS: *.* -XX:CICompilerCount=2\s
+
+  Scenario: CLOUD-1524, test JAVA_CORE_LIMIT and CONTAINER_CORE_LIMIT
+    When container is started with env
+      | variable             | value    |
+      | JAVA_CORE_LIMIT      | 3        |
+      | CONTAINER_CORE_LIMIT | 6        |
+    Then container log should match regex ^ *JAVA_OPTS: *.* -XX:ParallelGCThreads=6\s
```
maschmid: as the CONTAINER_CORE_LIMIT is not supposed to be set explicitly, the test should perhaps set it via the cpu docker argument

maschmid: IMHO the behavior should the opposite  (JAVA_CORE_LIMIT should take precedence, as CONTAINER_CORE_LIMIT is set from the actual cpu limit)

PR #66, opened 2017-08-18
CLOUD-1996: Added extra configuration to execution server image

rbarriuso:


os-bpmsuite-executionserver/added/launch/bpmsuite-executionserver.sh
```diff
@@ -66,6 +78,29 @@ function configure_controller_access {
     fi
 }
 
+function configure_router_access {
+    local routerService="${KIE_SERVER_ROUTER_SERVICE}"
+    routerService=${routerService^^}
+    routerService=${routerService//-/_}
+    # host
+    local kieServerRouterHost="${KIE_SERVER_ROUTER_HOST}"
+    if [ "${kieServerRouterHost}" = "" ]; then
```
psiroky: I think this won't work in case the `${kieServerRouterHost}` is actually empty. Bash will fail with something like `[: =: unary operator expected`. Using `if [ -z ${kieServerRouterHost} ];` (which checks is the length of the param is zero) should fix this.

rbarriuso: According to my tests it works, and actually the comparison `[ "${var}" = "" ]` is used in many other cct scripts. Nevertheless `-z` is also widely used. Perhaps we could create a JIRA to unify the comparison in all cct scripts. Wdyt?

psiroky: Ah, sorry, you are right. This would not work `[${kieServerRouterHost} = ""]`, but the current one will (the difference is the quotes).

Yes, I guess unifying this would not be a bad idea.

PR #67, opened 2017-08-21
Bdecoste cloud 1862

rimolive:
This PR will supersede [PR 22](https://github.com/jboss-openshift/cct_module/pull/22/)

### DIFF
--- a/os-datavirt/added/launch/teiid.sh
+++ b/os-datavirt/added/launch/teiid.sh
@@ -16,6 +16,9 @@ function prepareEnv() {
   unset DATAVIRT_USERS
   unset DATAVIRT_USER_PASSWORDS
   unset DATAVIRT_USER_GROUPS
+  unset JDBC_SECURITY_DOMAIN
+  unset ODBC_SECURITY_DOMAIN
+  unset ODATA_SECURITY_DOMAIN
 }
 
 function configure() {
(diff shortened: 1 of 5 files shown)


os-eap-launch/added/launch/security-domains.sh
```diff
@@ -47,23 +47,11 @@ function configure_security_domains() {
       fi
       security_domain="$security_domain >"
 
-      local options=$(compgen -v | grep -s "${domain_prefix}_MODULE_OPTION_")
+      local options=$(compgen -v | grep -sE "${domain_prefix}_MODULE_OPTION_[a-zA-Z]*(_[a-zA-Z]*)*_NAME")
```
rcernich: Would this be a little easier if it was _NAME_<id> as opposed to _<id>_NAME?  Generally, I think this is a more common pattern.

os-datavirt/added/launch/teiid.sh
```diff
@@ -16,6 +16,9 @@ function prepareEnv() {
   unset DATAVIRT_USERS
   unset DATAVIRT_USER_PASSWORDS
   unset DATAVIRT_USER_GROUPS
+  unset JDBC_SECURITY_DOMAIN
+  unset ODBC_SECURITY_DOMAIN
+  unset ODATA_SECURITY_DOMAIN
```
rcernich: Why are we defining separate domains for each transport?  Is this a common use case? In the event we do need this, would it be good to have a DEFAULT_SECURITY_DOMAIN that gets applied to all transports?

os-datavirt/added/launch/teiid.sh
```diff
@@ -139,6 +142,27 @@ function add_users(){
   done
 }
 
+function set_security_domains(){
+  if [ -n "$JDBC_SECURITY_DOMAIN" ]; then
+    sed -i "s|##JDBC_SECURITY_DOMAIN##|${JDBC_SECURITY_DOMAIN}|g" ${CONFIG_FILE}
+  else
+    sed -i "s|##JDBC_SECURITY_DOMAIN##|teiid-security|g" ${CONFIG_FILE}
```
rcernich: See comment above about default security domain.

os-eap-launch/added/launch/security-domains.sh
```diff
@@ -5,17 +5,65 @@ function prepareEnv() {
...
 
 function configureEnv() {
-  configure
+  configure_security_domains
+}
+
+function configure_security_domains() {
+  if [ -n "$SECURITY_DOMAINS" ]; then
+    for domain_prefix in $(echo $SECURITY_DOMAINS | sed "s/,/ /g"); do
+      local security_domain="<security-domain name=\"$domain_prefix\" cache-type=\"default\">"
```
rcernich: Would it be better to explicitly specify a NAME suffix for the domain name?  This type of thing caused problems with the DB prefix mapping stuff (where the prefix had to be of the form <service-name>_<db-type>).

os-datavirt/added/launch/teiid.sh
rimolive: I think this discussion falls on generalizing transports configuration, where the user can choose what transports to use.

rcernich: Understood, but the ability to specify a common default should make things easier for most use cases, e.g.
```
DEFAULT_SECURITY_DOMAIN=${DEFAULT_SECURITY_DOMAIN:-teiid-security}
JDBC_SECURITY_DOMAIN=${JDBC_SECURITY_DOMAIN:-${DEFAULT_SECURITY_DOMAIN}}
...
```

rcernich:
@rimolive, looks like you need to rebase.

rcernich:
Hey @rimolive, could you please add a variable to configure the default security domain as suggested in my comment above?

spolti:
This PR looks a little bit old, can we close this?

PR #74, opened 2017-08-30
[CLOUD-1867] update datavirt-app-config secret to use new src location

osnipso:
Change tests to rely on new default value for APP_DATADIR (== data)


tests/features/datavirt/datavirt_s2i.feature
```diff
@@ -30,7 +30,6 @@ Feature: Openshift JDV s2i tests
       | EXCEL_CONNECTION_JNDI                | java:/excel-file                                                 |
       | EXCEL_PROPERTY_ParentDirectory       | /home/jboss/source/data/teiidfiles/excelFiles/                   |
       | EXCEL_PROPERTY_AllowParentPaths      | true                                                             |
-      | APP_DATADIR                          | data                                                             |
```
rcernich: Should we add a test that checks for a non-default value for APP_DATADIR?

PR #76, opened 2017-09-01
CLOUD-2007 add PV partitioning to datagrid71

bdecoste:

### DIFF
--- a/os-jdg7-launch/added/openshift-launch.sh
+++ b/os-jdg7-launch/added/openshift-launch.sh
@@ -25,8 +25,31 @@ CONFIGURE_SCRIPTS=(
   /opt/run-java/proxy-options
 )
 
+function runServer() {
+  local instanceDir=$1
+
+  echo "Running $JBOSS_IMAGE_NAME image, version $JBOSS_IMAGE_VERSION"
+
+  exec $JBOSS_HOME/bin/standalone.sh -c clustered-openshift.xml -bmanagement 127.0.0.1 -Djboss.server.data.dir="$instanceDir" ${JBOSS_HA_ARGS} ${JAVA_PROXY_OPTIONS}
+}
+
+function init_data_dir() {
+  DATA_DIR="$1"
+  if [ -d "${JBOSS_HOME}/standalone/data" ]; then
+    cp -rf ${JBOSS_HOME}/standalone/data/* $DATA_DIR
+  fi
+}
+
 source $JBOSS_HOME/bin/launch/configure.sh
 
-echo "Running $JBOSS_IMAGE_NAME image, version $JBOSS_IMAGE_VERSION"
+if [ "${DATAGRID_SPLIT^^}" = "TRUE" ]; then
+  source /opt/partition/partitionPV.sh
 
-exec $JBOSS_HOME/bin/standalone.sh -c clustered-openshift.xml -bmanagement 127.0.0.1 ${JBOSS_HA_ARGS} ${JAVA_PROXY_OPTIONS}
+  DATA_DIR="${JBOSS_HOME}/standalone/partitioned_data"
+
+  partitionPV "${DATA_DIR}" "${DATAGRID_LOCK_TIMEOUT:-30}"
+else
+  echo "Running $JBOSS_IMAGE_NAME image, version $JBOSS_IMAGE_VERSION"
+
+  exec $JBOSS_HOME/bin/standalone.sh -c clustered-openshift.xml -bmanagement 127.0.0.1 ${JBOSS_HA_ARGS} ${JAVA_PROXY_OPTIONS}
+fi
--- a/os-partition/added/partitionPV.sh
+++ b/os-partition/added/partitionPV.sh
@@ -48,6 +48,11 @@ function partitionPV() {
           SERVER_DATA_DIR="${INSTANCE_DIR}/serverData"
           mkdir -p "${SERVER_DATA_DIR}"
 
+          if [ ! -f "${SERVER_DATA_DIR}/../data_initialized" ]; then
+            init_data_dir ${SERVER_DATA_DIR}
+            touch "${SERVER_DATA_DIR}/../data_initialized"
+          fi
+
           runServer "${SERVER_DATA_DIR}" &
 
           PID=$!
--- a/os-amq-launch/added/drain.sh
+++ b/os-amq-launch/added/drain.sh
@@ -6,7 +6,7 @@ if [ "${SCRIPT_DEBUG}" = "true" ] ; then
 fi
 
 source $AMQ_HOME/bin/configure.sh
-source $AMQ_HOME/bin/partitionPV.sh
+source /opt/partition/partitionPV.sh
 source $AMQ_HOME/bin/drainClasspath.sh
 source /usr/local/dynamic-resources/dynamic_resources.sh
 
--- a/os-amq-launch/added/launch.sh
+++ b/os-amq-launch/added/launch.sh
@@ -6,7 +6,7 @@ if [ "${SCRIPT_DEBUG}" = "true" ] ; then
 fi
 
 source $AMQ_HOME/bin/configure.sh
-source $AMQ_HOME/bin/partitionPV.sh
+source /opt/partition/partitionPV.sh
 source /usr/local/dynamic-resources/dynamic_resources.sh
 
 ACTIVEMQ_OPTS="$(adjust_java_options ${ACTIVEMQ_OPTS})"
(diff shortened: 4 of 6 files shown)


rcernich:
Do we allow the user to specify DATA_DIR manually already?  If so, we need to incorporate that into the scripts.

@knrc, do you mind reviewing the partition script changes?

knrc:
@rcernich  This is a copy/paste of the original scripts so we need to look at refactoring to make them generic, I can take a look at that.

os-jdg7-launch/added/partitionPV.sh
```diff
@@ -0,0 +1,209 @@
...
+
+              if [ $LOCK_STATUS -eq 0 ] ; then
+                echo "Successfully locked directory: ($INSTANCE_DIR)"
+                MIGRATED=true
+
+                flock -u $MIGRATING_FD
+
+                SERVER_DATA_DIR="${INSTANCE_DIR}"
+
+                PID=$!
```
knrc: This doesn't make any sense, nothing is being migrated here.  Are we intending to migrate something?

os-jdg7-launch/added/partitionPV.sh
```diff
@@ -0,0 +1,209 @@
...
+      if [ "$MIGRATED" = "false" ] ; then
+        echo "Finished Migration Check cycle, pausing for ${MIGRATION_PAUSE} seconds before resuming"
+        COUNT=1
+        sleep "${MIGRATION_PAUSE}"
+      else
+        MIGRATED=false
+      fi
+    fi
+  done
+}
```
knrc: This is a complete copy of the a-mq script with two minor changes
- SERVER_DATA_DIR has been changed, it should be ${INSTANCE_DIR}/serverData
- the runMigration part has been removed.

I suspect the original script can be used as-is which was certainly my intention, is there any reason why this was copied?

knrc:
I also suspect the data directory should be somewhere else, not under data, since for EAP installations that could potentially contain initialised files from the S2I build.  We would need to introduce support for initialisation in the generic script so those files could be copied when a new directory is created.

os-jdg7-launch/added/partitionPV.sh
bdecoste: My intention was to remove the runMigration() call which contains the AMQ-specific logic as JDG just cares about the file system partition. So I copied the AMQ drain.sh to JDG migratePV.sh and removed the AMQ specific call. The JDG drain pod calls migratePV.sh not drain.sh (see here https://github.com/jboss-openshift/application-templates/pull/335/files).

Then as you indicated I only changed the SERVER_DATA_DIR value in partitionPV.sh as I couldn't figure out what serverData was and assumed it was something AMQ specific. And removed the call to runMigration() since that's the AMQ specific call that now doesn't exist for JDG. 

Does JDG not need the drain pod and the migratePV() call at all? If that's the case then we should be able to use the existing AMQ logic and eliminate the JDG drain pod completely.

os-partition/added/partitionPV.sh
```diff
@@ -48,6 +50,10 @@ function partitionPV() {
           SERVER_DATA_DIR="${INSTANCE_DIR}/serverData"
           mkdir -p "${SERVER_DATA_DIR}"
 
+          if [ -d "$S2I_DATA_DIR" ]; then
+            cp -rf ${S2I_DATA_DIR}/* ${SERVER_DATA_DIR}
+          fi
+
```
knrc: This should only run on the first instance, i.e. when we create the directory for the first time.  It should also be handled through a function, perhaps init_server_data or similar, so there are no product specific details within this script.  We would need to stub this out in the A-MQ script since I don't believe it has requirements for initialising the data dir.

os-partition/configure.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+SCRIPT_DIR=$(dirname $0)
+ADDED_DIR=${SCRIPT_DIR}/added
+
+if [ -n "$AMQ_HOME" ]; then
+  cp -p ${ADDED_DIR}/partitionPV.sh $AMQ_HOME/bin/
+else
+  cp -p ${ADDED_DIR}/partitionPV.sh $JBOSS_HOME/bin/
```
knrc: should this be bin/launch rather than bin?

os-jdg7-launch/added/openshift-launch.sh
```diff
@@ -25,8 +25,25 @@ CONFIGURE_SCRIPTS=(
...
 
+function runServer() {
+  local instanceDir=$1
+
+  echo "Running $JBOSS_IMAGE_NAME image, version $JBOSS_IMAGE_VERSION"
+
+  exec $JBOSS_HOME/bin/standalone.sh -c clustered-openshift.xml -bmanagement 127.0.0.1 -Djboss.server.data.dir="$instanceDir" ${JBOSS_HA_ARGS} ${JAVA_PROXY_OPTIONS}
+}
+
 source $JBOSS_HOME/bin/launch/configure.sh
```
knrc: Define the init_data_dir function here

os-partition/added/partitionPV.sh
```diff
@@ -48,6 +48,11 @@ function partitionPV() {
           SERVER_DATA_DIR="${INSTANCE_DIR}/serverData"
           mkdir -p "${SERVER_DATA_DIR}"
 
+          if [ -f /opt/partition/copyDataDir.sh ]; then
+            source /opt/partition/copyDataDir.sh
+            copy_data_dir ${SERVER_DATA_DIR}
+          fi
+
```
knrc: This should be a function call just like runServer, defined in the script sourcing the partitionPV.sh.

os-jdg7-launch/added/openshift-launch.sh
knrc: It would need an equivalent in the A-MQ script, albeit an empty function.

function init_data_dir() {
  // No initialisation needed
}

os-partition/added/partitionPV.sh
knrc: Once the function is defined in the sourcing script this could be

init_data_dir "${SERVER_DATA_DIR}"

os-jdg7-launch/added/openshift-launch.sh
```diff
@@ -25,8 +25,35 @@ CONFIGURE_SCRIPTS=(
...
 
-echo "Running $JBOSS_IMAGE_NAME image, version $JBOSS_IMAGE_VERSION"
+if [ "$DATAGRID_SPLIT" = "true" ]; then
+  source /opt/partition/partitionPV.sh
 
-exec $JBOSS_HOME/bin/standalone.sh -c clustered-openshift.xml -bmanagement 127.0.0.1 ${JBOSS_HA_ARGS} ${JAVA_PROXY_OPTIONS}
+  DATA_DIR="${JBOSS_HOME}/standalone/data"
```
knrc: The assemble script will copy the contents into standalone/data so it may be better for this to be a different directory, perhaps standalone/partition_data or a better name?

Apologies for not catching this before.

os-partition/added/partitionPV.sh
```diff
@@ -48,6 +48,8 @@ function partitionPV() {
           SERVER_DATA_DIR="${INSTANCE_DIR}/serverData"
           mkdir -p "${SERVER_DATA_DIR}"
 
+          init_data_dir ${SERVER_DATA_DIR}
```
knrc: The check for initialised data is a good idea, would it make more sense to move the check here?  It would not then be needed in the init_data_dir function.

if [ ! -f "${SERVER_DATA_DIR}/../data_initialized" ]; then
      init_data_dir "${SERVER_DATA_DIR}"
      touch "${SERVER_DATA_DIR}/../data_initialized"
fi

os-jdg7-launch/added/openshift-launch.sh
```diff
@@ -25,8 +25,35 @@ CONFIGURE_SCRIPTS=(
...
+  echo "Running $JBOSS_IMAGE_NAME image, version $JBOSS_IMAGE_VERSION"
+
+  exec $JBOSS_HOME/bin/standalone.sh -c clustered-openshift.xml -bmanagement 127.0.0.1 -Djboss.server.data.dir="$instanceDir" ${JBOSS_HA_ARGS} ${JAVA_PROXY_OPTIONS}
+}
+
+function init_data_dir() {
+  DATA_DIR="$1"
+  if [ -d /tmp/src/data ]; then
+    if [ ! -f "${DATA_DIR}/../data_initialized" ]; then
+      cp -rf /tmp/src/data/* $DATA_DIR
```
knrc: If we assume the assemble script has done its job then it may be better to copy standalone/data into "${DATA_DIR}", this could also catch any changes an extension may choose to make.

@rcernich Would we support extensions seeding information into the data directory?  I suspect your work allows this but would like to confirm.

rcernich:
rebased and merged, with added module.yaml for new package

PR #82, opened 2017-09-06
[CLOUD-1925] [EAP71] Set activemq thread pool system properties

kyguy:
JIRA link: https://issues.jboss.org/browse/CLOUD-1925

Currently, the JVM is not cgroup aware and cannot be trusted to generate default values for Artemis
threads pool arguments. The thread pool arguments must respect the limits of its containerized environment, and therefore are calculated using the original methods provided by Artemis and the resources reported by cgroups.

### DIFF
--- a/os-eap7-launch/added/launch/messaging.sh
+++ b/os-eap7-launch/added/launch/messaging.sh
@@ -62,6 +62,7 @@ function prepareEnv() {
 function configure() {
   inject_brokers
   configure_mq
+  configure_thread_pool
 }
 
 function configure_mq_destinations() {
@@ -99,6 +100,20 @@ function configure_mq() {
 
     sed -i "s|<!-- ##MESSAGING_SUBSYSTEM_CONFIG## -->|${activemq_subsystem%$'\n'}|" "${CONFIG_FILE}"
     sed -i 's|<!-- ##MESSAGING_PORTS## -->|<socket-binding name="messaging" port="5445"/><socket-binding name="messaging-throughput" port="5455"/>|' "${CONFIG_FILE}"
+  fi
+}
+
+# Currently, the JVM is not cgroup aware and cannot be trusted to generate default values for
+# threads pool args. Therefore, if there are resource limits specifed by the container, this function
+# will configure the thread pool args using cgroups and the formulae provied by https://github.com/apache/activemq-artemis/blob/master/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/client/ActiveMQClient.java
+function configure_thread_pool() {
+  source /opt/run-java/container-limits
+  if [ -n "$CONTAINER_CORE_LIMIT" ]; then
+    local mtp=$(expr 8 \* $CONTAINER_CORE_LIMIT) # max thread pool size
+    local ctp=5                                  # core thread pool size
+    JBOSS_MESSAGING_ARGS="${JBOSS_MESSAGING_ARGS}
+    -Dactivemq.artemis.client.global.thread.pool.max.size=$mtp
+    -Dactivemq.artemis.client.global.scheduled.thread.pool.core.size=$ctp"
   fi
 }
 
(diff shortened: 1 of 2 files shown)


os-eap7-launch/added/launch/messaging.sh
```diff
@@ -102,6 +103,20 @@ function configure_mq() {
...
+# threads pool args. Therefore, if there are resource limits specifed by the container, this function
+# will configure the thread pool args using cgroups and the formulae provied by https://github.com/apache/activemq-artemis/blob/master/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/client/ActiveMQClient.java
+function configure_thread_pool() {
+  source /opt/run-java/container-limits
+  if [ -n "$CONTAINER_CORE_LIMIT" ]; then
+    local mtp=$(expr 8 \* $CONTAINER_CORE_LIMIT) # max thread pool size
+    local ctp=5                                  # core thread pool size
+    JBOSS_MESSAGING_ARGS="${JBOSS_MESSAGING_ARGS}
+    -Dactivemq.artemis.client.global.thread.pool.max.size=$mtp
+    -Dactivemq.artemis.client.global.scheduled.thread.pool.core.size=$ctp"
```
tremes: I would like to ask couple of newbie questions.:-) How can user set these values when there's no `$CONTAINER_CORE_LIMIT`? 
And second - this is passed as arguments right? Wouldn't be more clear to define this directly in EAP standalone configuration?

kyguy: The variable `$CONTAINER_CORE_LIMIT` only exists if the container's resources are restricted. Currently, we calculate the container's core limit using this function https://github.com/jboss-openshift/cct_module/blob/8cec4cc294d8fa8e3658390737c7a75f52cbeac0/os-java-run/added/container-limits#L22-L37 taking the ceiling of cpu_quota/cpu_period. We can make sure the `$CONTAINER_CORE_LIMIT` variable is assigned a value by assigning the container arguments cpu_quota and cpu_period to values that would make ceiling(cpu_quota/cpu_period) > 0.
```bash
docker run -it --cpu-quota=100000 --cpu-period=100000 --rm <IMAGE_ID> bash
```
will give a `$CONTAINER_CORE_LIMIT` of 1. If `$CONTAINER_CORE_LIMIT` is not defined, then the container has no resource limits and we can allow the JVM to automatically calculate default values for the thread pool arguments.

A user can manually set the thread pools by adding the parameters at the end of the exec statement in the openshift-launch script (https://github.com/jboss-openshift/cct_module/blob/master/os-eap7-launch/added/openshift-launch.sh#L42)
```bash
exec $JBOSS_HOME/bin/standalone.sh -c standalone-openshift.xml -bmanagement 127.0.0.1 ${JAVA_PROXY_OPTIONS} ${JBOSS_HA_ARGS} ${JBOSS_MESSAGING_ARGS} -Dactivemq.artemis.client.global.thread.pool.max.size=8 -Dactivemq.artemis.client.global.scheduled.thread.pool.core.size=5
```
Yes, activemq.artemis.client.global.thread.pool.max.size and activemq.artemis.client.global.scheduled.thread.pool.core.size are passed as arguments to the $JBOSS_HOME/bin/standalone.sh above as part of the `$JBOSS_MESSAGING_ARGS`. I believe these arguments are split into categories (e.g. `$JAVA_PROXY_OPTIONS`, `$JBOSS_HA_ARGS`, `$JBOSS_MESSAGING_ARGS`) in order to keep the arguments organized as the number of arguments increase. Although, I do agree with you, it would be nice to have all these in one place like the configuration file :). I think the configuration file is reserved for arguments that are not classified into the defined categories

Let me know if you have any other questions :)

P.S.
On a side note, it looks like we might need to revisit to core limit calculations. For one reason we can make `$CONTAINER_CORE_LIMIT` greater than the number of cores on the actual host
by messing with the cpu_quota and cpu_period. Another reason is that `$CONTAINER_CORE_LIMIT` appears to be unaffected by the setting of cpuset-cpus and cpus container arguments. Also it appear that the standard for defining the cpu limit is changing to the --cpus argument for docker 1.13 (https://docs.docker.com/engine/admin/resource_constraints/#cpu) Anyways, it would be nice to have a uniformly defined core limit across a container

tremes: Thanks for comprehensive explanation. Makes sense.

PR #83, opened 2017-09-07
CLOUD-1554 enable toggling of 32 vs 64 bit JVM

bdecoste:

### DIFF
--- a/os-java-run/added/java-alternatives.sh
+++ b/os-java-run/added/java-alternatives.sh
@@ -1,0 +1,9 @@
+set_java_alternatives() {
+  if [ "${USE_32_BIT_JVM^^}" = "TRUE" ]; then
+    JavaHome32=`alternatives --display java | grep family | grep i386/ | awk '{print $1}' | sed 's|/jre/bin/java||'`
+    echo "$JavaHome32"
+  else
+    echo "$JAVA_HOME"
+  fi
+}
+
--- a/dynamic-resources/dynamic_resources.sh
+++ b/dynamic-resources/dynamic_resources.sh
@@ -17,7 +17,14 @@ function source_java_run_scripts() {
     source "${java_scripts_dir}/java-default-options"
 }
 
+function set_java_arch(){
+    source /opt/run-java/java-alternatives.sh
+
+    export JAVA_HOME="$(set_java_alternatives)"
+}
+
 source_java_run_scripts
+set_java_arch
 
 # deprecated, left for backward compatibility
 function get_heap_size {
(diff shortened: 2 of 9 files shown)


java-alternatives/run.sh
```diff
@@ -56,3 +56,10 @@ set_java_alternatives() {
 }
 
 set_java_alternatives
+
+SCRIPT_DIR=$(dirname $0)
+ADDED_DIR=${SCRIPT_DIR}/added
+
+mkdir -p ${JBOSS_HOME}/bin/launch
+cp -r ${ADDED_DIR}/launch/* ${JBOSS_HOME}/bin/launch
+
```
knrc: Would it not be better to include this within the os-java-run scripts so it is available to all java images?

os-eap7-openshift/added/standalone.conf
```diff
@@ -12,3 +12,8 @@ fi
 
 # Append to JAVA_OPTS. Necessary to prevent some values being omitted if JAVA_OPTS is defined directly
 JAVA_OPTS="$JAVA_OPTS $JAVA_OPTS_APPEND"
+
+source $JBOSS_HOME/bin/launch/java-alternatives.sh
+
+export JAVA_HOME="$(set_java_alternatives)"
+
```
knrc: This appears to be EAP only but I don't see the yaml change to pull in the java-alternatives module, is there another PR associated with this?

bdecoste: I have just done EAP7 initially to get it reviewed. The java-alternatives module is already in the jboss-eap-7-openshift-image image.yaml, but I will move to os-java-run and update the corresponding files for all images (e.g. standalone.conf). I was really hoping to find a way where I didn't have to touch files specific to each image but that seems like the only way.

bdecoste: This is the PR that installs the 32bit JVM: https://github.com/jboss-container-images/jboss-openjdk-image/pull/7

knrc: +1 I found that yesterday when reviewing this PR

os-java-run/added/java-alternatives.sh
```diff
@@ -0,0 +1,9 @@
+set_java_alternatives() {
+  if [ "${USE_32_BIT_JVM^^}" = "TRUE" ]; then
```
rcernich: I think the goal of the jira was to automatically set this if the container size was less than a certain size.  Furthermore, there should probably be a way to disable using 32bit jvm if an user's application requires a native library that's unavailable (e.g. apr).

knrc: +1 on setting this automatically, there's an example of this in the jenkins image

https://github.com/openshift/jenkins/blob/master/2/contrib/s2i/run#L126-L130

I'm not sure if it will be possible to determine reliably whether the libraries are present or not, they may be too far.

rcernich: I don't think we need to detect the presence of any native libs that may be required.  We just need a way for the user to disable using 32bit if they choose.  In other words, if they don't disable this feature, we use 32bit jvm if the container is not big enough for 64bit.

PR #96, opened 2017-09-20
[CLOUD-2045] remove two unusual files

jmtd:
https://issues.jboss.org/browse/CLOUD-2045

Two spurious files we don't need.


goldmann:
LGTM, just wanted to make sure that we can start merging things with @rcernich

PR #103, opened 2017-09-26
CLOUD-2049 deterministic execution ordering of valves

bdecoste:

### DIFF
--- a/os-eap-launch/added/launch/access_log_valve.sh
+++ b/os-eap-launch/added/launch/access_log_valve.sh
@@ -35,16 +35,14 @@ function configure() {
 }
 
 function configure_access_log_valve() {
-    EAP6_VALVE="<valve name=\"accessLog\" module=\"org.jboss.openshift\" class-name=\"org.jboss.openshift.valves.StdoutAccessLogValve\">\n              \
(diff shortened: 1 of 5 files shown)


maschmid:
Does the overlay mean we are replacing any jboss-web.xml that might be in the deployment? 

It does seem too hacky to me.

rcernich:
Good question.  I believe it is merged with the existing content, but I'd need to check.  The KB suggests this as the mechanism for addressing the issue globally.  In either case, it seems we need to determine the side effects of this approach.

bdecoste:
The global jboss-web.xml doesn't replace the individual application's jboss-web.xml. It adds the valves to all deployments.

maschmid:
I have tried it with a simple war with WEB-INF/jboss-web.xml containing:
<jboss-web><context-root>/foobar</context-root></jboss-web>

on the new image the context root seems to be ignored, so it does seem the overlay overrides the deployments' jboss-web.xml

bstansberry:
A deployment-overlay replaces the specified file; there is no merging of content. So putting one in for jboss-web.xml means whatever was already in the app is lost.

PR #111, opened 2017-10-02
Separated openshift-layer module to two

dbecvarik:
Signed-off-by: David Becvarik <dbecvari@redhat.com>


rcernich:
I think we should have tried to determine why this was happening so we could address the root of the problem instead of the symptom.  Eventually, we'll be merging modules so they are coarser grained and splitting them apart further goes against that goal.  It also requires modifying all the images to add the new module.

I'll see if I can find the root of the problem.

rcernich:
If this was causing image builds to fail, I wonder if there isn't a problem with the generator in concreate.  All the configure_passwd.sh script does is chmod /etc/passwd.  The passwd file is not used or referenced in any other module.

rcernich:
I don't think this change is bad and is probably something we want to add to our base rhel image.  Furthermore, the openshift-layers module is specific to EAP, so I think it's good that they've been separated.

dbecvarik:
the issue was not configure_passwd but configure_layers one failed AMQ and webserver builds. Theese images need to run configura_passwd but cannot run configure layers.

rcernich:
Ahh, so those images had previously only been running the one script and not both.  Makes sense now.  Thanks @dbecvarik.

PR #137, opened 2017-10-24
[CLOUD-2068] make refresh-interval for timer data-store configurable

psiroky:
The refresh-interval is not supported by EAP 6.4.x, so can not be added
by default.

Signed-off-by: Petr Široký <psiroky@redhat.com>

Second try, hopefully compatible with all the images now. Tried with EAP 6.4 and it works (as long as the env. var `TIMER_SERVICE_DATA_STORE_REFRESH_INTERVAL` is not set, which it should not for EAP 6.4 images as the refresh interval is not supported there).

### DIFF
--- a/os-eap-launch/added/launch/datasource-common.sh
+++ b/os-eap-launch/added/launch/datasource-common.sh
@@ -186,8 +186,9 @@ function generate_datasource_common() {
   fi
 
   if [ -n "$TIMER_SERVICE_DATA_STORE" -a "$TIMER_SERVICE_DATA_STORE" = "${service_name}" ]; then
+    local refreshInterval=${TIMER_SERVICE_DATA_STORE_REFRESH_INTERVAL:--1}
     inject_timer_service ${pool_name}_ds
-    inject_datastore $pool_name $jndi_name $driver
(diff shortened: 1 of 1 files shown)


knrc:
This would still be an issue if the environment variable was set, it's better to have something specific in the eap7/eap6 datasource.sh files that can then be consumed by the common script.  I'll propose a solution later today and send the details so you can give your thoughts.

psiroky:
I am fine with whatever lets me set the `refresh-interval` using a env var.

psiroky:
Do we have any update on this?

knrc:
Apologies, I was pulled in to other firefights and this slipped down the list.  I'll handle this later today.

knrc:
@psiroky The following commit shows what I was suggesting, I've also added tests into the commit to show how those should work

https://github.com/knrc/cct_module/commit/913123071111467a702fff6f956e9d248c2be1b5

Note I also had to add this to the SSO modules but I've created another JIRA to get that duplication addressed.

psiroky:
Looks good, thanks @knrc! Closing this one.

PR #148, opened 2017-11-10
[CLOUD-2139] Upgrade Maven to 3.5

kyguy:

https://issues.jboss.org/browse/CLOUD-2139

### DIFF
--- a/jboss-maven/configure.sh
+++ b/jboss-maven/configure.sh
@@ -1,12 +1,21 @@
 #!/bin/sh
-# Add default Maven settings with Red Hat/JBoss repositories
+# Configure module
 set -e
 
 SCRIPT_DIR=$(dirname $0)
 ADDED_DIR=${SCRIPT_DIR}/added
 
+chown -R jboss:root $SCRIPT_DIR
+chmod -R ug+rwX $SCRIPT_DIR
+
+# Add default Maven settings with Red Hat/JBoss repositories
 mkdir -p $HOME/.m2
 cp -p ${ADDED_DIR}/jboss-settings.xml $HOME/.m2/settings.xml
+
+# Add s2i scripts
+mkdir -p /usr/local/s2i
+cp -p ${ADDED_DIR}/s2i/scl-enable-maven /usr/local/s2i/
+chmod ug+x /usr/local/s2i/scl-enable-maven
 
 chown -R jboss:root $HOME
 chmod -R g+rwX $HOME
--- a/jboss-maven/module.yaml
+++ b/jboss-maven/module.yaml
@@ -1,7 +1,85 @@
 schema_version: 1
 name: jboss-maven
 version: '1.0'
-description: Legacy jboss-maven script package.
+description: Maven module
+labels:
+    - name: io.fabric8.s2i.version.maven
+      value: "3.5"
+envs:
+    - name: MAVEN_VERSION
+      value: "3.5"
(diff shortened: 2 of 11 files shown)


jboss-maven/configure.sh
```diff
@@ -1,12 +1,21 @@
...
+chmod -R g+rwX $SCRIPT_DIR
+
+# Add default Maven settings with Red Hat/JBoss repositories
 mkdir -p $HOME/.m2
 cp -p ${ADDED_DIR}/jboss-settings.xml $HOME/.m2/settings.xml
 
+# Add s2i scripts
+chmod g+x $ADDED_DIR/s2i/*
+mkdir -p /usr/local/s2i
+cp -p ${ADDED_DIR}/s2i/* /usr/local/s2i/
```
dbecvarik: I would prefer to set permission here, instead of setting them in $ADDED_DIR and then copy with preserve - its harder to read. But its very minor.

jboss-maven/module.yaml
```diff
@@ -1,7 +1,85 @@
...
+      description: "Maven repository snapshots enabled"
+    - name: "MAVEN_REPO_SNAPSHOTS_UPDATE_POLICY"
+      example: "always"
+      description: "Maven repository snapshots update policy"
 execute:
 - script: configure.sh
-  user: '185'
+packages:
+  repositories:
+          - jboss-rhscl
```
dbecvarik: inconsistent indentation - should be same as - rh-maven35

jboss-maven/module.yaml
```diff
@@ -1,7 +1,85 @@
 schema_version: 1
 name: jboss-maven
 version: '1.0'
-description: Legacy jboss-maven script package.
+description: Maven module
+labels:
+    - name: io.fabric8.s2i.version.maven
+      value: "3.5"
```
rcernich: Do we need a more precise version?  This is truncated compared to the old version number we were setting.

kyguy: From what I understand, I don't think we need a more precise version

With or without this label, the images build and pass all tests; I believe it is there for cosmetic reasons. I brought this label over from the redhat-openjdk-18-openshift image: ( https://github.com/jboss-container-images/redhat-openjdk-18-openshift-image/blob/openjdk18-dev/image.yaml#L9-L10 ) and changed its value to match that of the environment variable, MAVEN_VERSION

If we were to define a more precise version, wouldn't we need to update it more frequently? Also, how do we know which version to choose? As I understand, concreate pulls the latest version of the defined package, rh-maven35, specified by the software collections repo, but we don't know the exact version specified for sure until the image is built. If we leave the label with just the major version, we will always be correct

rcernich: I can go along with that, but the same argument held true previously, so... @goldmann, what's your take?

goldmann: It should point only to major.minor IMHO. So exactly as it is proposed now.

goldmann:
Obviously, other question is whether this label is required or not.

rcernich:
Yeah, but it's already there, so...  I'm not going to ask for it to be removed as part of this change.  We can create a separate issue if we think that is important.

goldmann:
Of course, this was just a random comment. Once we release modules trully independently we will be able to do it as a major release.

rcernich:
manually rebased and merged

PR #161, opened 2017-12-05
CLOUD-2211 Updates for JDG 7.1.1

slaskawi:
https://issues.jboss.org/browse/CLOUD-2211

This PR includes JDG 7.1.1 Test fix and `Encryption` protocol fix. 

Depending what other products will do, we might want to collapse the product `if` statement to JGroups version comparison (>= JGroups 4 use `SYM_ENCRYPT` for example). However this requires a longer thought and making sure we don't break anything. I might file a separate JIRA for it but for the time being (chasing by the release date), I would stick to a simple `or` statement.

### DIFF
--- a/os-eap-launch/added/launch/jgroups.sh
+++ b/os-eap-launch/added/launch/jgroups.sh
@@ -36,7 +36,8 @@ function configure_jgroups_encryption() {
       product=`read_product_name`
       version=`read_product_version`
 
-      if [ "$product" == "Red Hat JBoss Enterprise Application Platform" -a "$(version_compare $version '6.4.4.GA')" == "newer" ]; then
+      if [ "$product" == "Red Hat JBoss Enterprise Application Platform" -a "$(version_compare $version '6.4.4.GA')" == "newer" ] ||
+         [ "$product" == "Red Hat JBoss Data Grid" -a "$(version_compare $version '7.1.0')" == "newer" ]; then
         # For new JGroups we need to use SYM_ENCRYPT protocol
         jgroups_encrypt="\
           <protocol type=\"SYM_ENCRYPT\">\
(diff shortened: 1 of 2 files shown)


os-eap-launch/added/launch/jgroups.sh
```diff
@@ -36,7 +36,8 @@ function configure_jgroups_encryption() {
       product=`read_product_name`
       version=`read_product_version`
 
-      if [ "$product" == "Red Hat JBoss Enterprise Application Platform" -a "$(version_compare $version '6.4.4.GA')" == "newer" ]; then
+      if [ "$product" == "Red Hat JBoss Enterprise Application Platform" -a "$(version_compare $version '6.4.4.GA')" == "newer" ] ||
```
slaskawi: @goldmann Asked for clarification why it did work in JDG 7.0.x.

It turned out that it worked because of a bug. We didn't change the product name in `version.txt`:

```
JDG 7.0.x:
cat /opt/datagrid/version.txt
Red Hat JBoss Enterprise Application Platform - Version 7.0.4.GA

JDG 7.1.1:
cat /opt/datagrid/version.txt 
Red Hat JBoss Data Grid - Version 7.1.1
```

rcernich: I think this is OK, but I'm concerned that we've now got JDG specific code in a common EAP module.  Hindsight being 20/20, perhaps we should have had separate JGroups configuration scripts for EAP 6 and 7.

slaskawi: As I mentioned in PR's description, I believe checking JGroups version might be a much better idea. I already prepared a Pull Request for this: https://github.com/jboss-openshift/cct_module/pull/164.

If you agree with me, I prefer to close this PR (as it's a defensive implementation) and pursue https://github.com/jboss-openshift/cct_module/pull/164.

goldmann:
Closing, so this does not interfere with other PR.

PR #167, opened 2017-12-08
[CLOUD-2221] configure KieLoginModule only for Business Central

psiroky:

This is definitely not an ideal solution. We should probably have generic set of common functions/modules and then specify which of them would be executed for the specific deliverables. But until that is in place this should be a sufficient workaround.

### DIFF
--- a/tests/features/bpmsuite/businesscentral/configuration.feature
+++ b/tests/features/bpmsuite/businesscentral/configuration.feature
@@ -1,0 +1,8 @@
+@jboss-bpmsuite-7/bpmsuite70-businesscentral-openshift @jboss-bpmsuite-7/bpmsuite70-businesscentral-monitoring-openshift
+Feature: Business Central configuration tests
+
+  # https://issues.jboss.org/browse/CLOUD-2221
+  Scenario: Check KieLoginModule is configured
+      When container is ready
+
+      Then file /opt/eap/standalone/configuration/standalone-openshift.xml should contain <login-module code="org.kie.security.jaas.KieLoginModule"
--- a/tests/features/rhdm/decisioncentral/configuration.feature
+++ b/tests/features/rhdm/decisioncentral/configuration.feature
@@ -1,0 +1,8 @@
+@rhdm-7/rhdm70-decisioncentral-openshift
+Feature: Decision Central configuration tests
+
+  # https://issues.jboss.org/browse/CLOUD-2221
+  Scenario: Check KieLoginModule is _not_ configured
+      When container is ready
+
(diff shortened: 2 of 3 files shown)


psiroky:
CC @goldmann

psiroky:
@spolti could you please take a look and review this?

spolti:
@psiroky  wouldn't businesscentral-monitoring need this too?

psiroky:
@spolti good point about the businesscentral-monitoring - that will likely need it too (or at least it was the current behaviour, so better leave it as is).

I don't see any bmpsuite tests (yay) to start from, but will take a look at adding some based on tests from other modules.

spolti:
I created a jira to add those tests: CLOUD-2213
I added a couple of basic tests, @goldmann @psiroky do you think we can get the tests merged?

psiroky:
I have approved the PRs for CLOUD-2213. I am afraid I can't do much more at this point.

psiroky:
@spolti tests added. Could you please check again?

spolti:
You added tests for rhdm but in the scripts you check only for businesscentral, is this configuration needed on decision manager too?

psiroky:
@spolti the tests are making sure the custom login-module is _not_ configured as it actually breaks the Decision Central (see the linked JIRA).

tests/features/bpmsuite/businesscentral/configuration.feature
```diff
@@ -0,0 +1,11 @@
+@jboss-bpmsuite-7/bpmsuite70-businesscentral-openshift @jboss-bpmsuite-7/bpmsuite70-businesscentral-monitoring-openshift
+Feature: Business Central configuration tests
+
+  # https://issues.jboss.org/browse/CLOUD-2221
+  Scenario: Check KieLoginModule is configured
+  Given XML namespaces
+        | prefix | url                           |
```
spolti: It is not necessary as you aren't using xml namespace on the tests.

tests/features/rhdm/decisioncentral/configuration.feature
```diff
@@ -0,0 +1,11 @@
+@rhdm-7/rhdm70-decisioncentral-openshift
+Feature: Decision Central configuration tests
+
+  # https://issues.jboss.org/browse/CLOUD-2221
+  Scenario: Check KieLoginModule is _not_ configured
+  Given XML namespaces
```
spolti: same here.

tests/features/bpmsuite/businesscentral/configuration.feature
psiroky: Ah, good point, thanks. I was trying to use the xpath way, but that's actually not necessary for this simple case. Will remove that.

PR #160, opened 2017-12-04
CLOUD-2210: Add transaction recovery support for EAP 6.4 and EAP 7.0

knrc:
Thanks for submitting your Pull Request!

Please make sure your PR meets following requirements:

- [ ] Pull Request title is properly formatted: `[CLOUD-XYA] Subject`
- [ ] Pull Request contains link to the JIRA issue
- [ ] Pull Request contains description of the issue
- [ ] Pull request does not include fixes for other issues than the main ticket
- [ ] Attached commits represent unit of work and are properly formatted
- [ ] You have read and agreed to the Developer Certificate of Origin (DCO) (see `CONTRIBUTING.md`)
- [ ] Every commit contains `Signed-off-by: Your Name <yourname@redhat.com>` - use `git commit -s`

PR #171, opened 2017-12-13
[CLOUD-2223] table prefix for jdbc object store table defined by default

ochaloup:
When the object store is used as storage for transactions the table where txn data are saved should be unique for each pod. That's ensured by adding prefix to the table with txn records which is - `jbosststxtable`.

https://issues.jboss.org/browse/CLOUD-2223


ochaloup:
the fix was already provided by https://github.com/jboss-openshift/cct_module/pull/160

PR #211, opened 2018-03-14
[CLOUD-2379] Allow override LAUNCH_DIR

ruromero:
Part of the fix for https://issues.jboss.org/browse/RHBA-530
Fixes https://issues.jboss.org/browse/CLOUD-2379
Allows the possibility for non-eap images to decide where to copy the scripts

### DIFF
--- a/os-logging/module.yaml
+++ b/os-logging/module.yaml
@@ -4,4 +4,3 @@ version: '1.0'
 description: os-logging script package.
 execute:
 - script: configure.sh
-  user: '185'
--- a/os-logging/configure.sh
+++ b/os-logging/configure.sh
@@ -12,8 +12,10 @@ else
   BIN_HOME="$JBOSS_HOME"
 fi
 
-mkdir -p ${BIN_HOME}/bin/launch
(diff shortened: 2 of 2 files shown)


ruromero:
I have removed the user tag because for doing a `chown jboss:root` it must be root

os-logging/module.yaml
```diff
@@ -4,4 +4,3 @@ version: '1.0'
 description: os-logging script package.
 execute:
 - script: configure.sh
-  user: '185'
```
rcernich: Why was this removed?

ruromero: @rcernich because this script must be run as `root` as it executes `chown` command

rcernich: It was doing that before.  Is this because LAUNCH_DIR is not owned by jboss user?

ruromero: @rcernich I don't know how come it works for other dependencies but `chown jboss:root <file>` can't work without root permissions UNLESS the ownership was already `jboss:root` meaning that nothing was changed

PR #217, opened 2018-03-22
[CLOUD-2402] Add SSO_CLIENT

ruromero:
Fixes https://issues.jboss.org/browse/CLOUD-2402
The keycloak.sh script was not taking into consideration the SSO_CLIENT environment variable to set the "resource" as described in the [EAP 7.0 docs](https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.0/html-single/red_hat_jboss_enterprise_application_platform_for_openshift/#environment_variables_4)

Additionally
* Some useless spaces have been removed
* Some methods have been refactor for readability 

Depends on: https://github.com/jboss-openshift/cct_module/pull/211


os-eap-sso/added/keycloak.sh
```diff
@@ -157,17 +158,17 @@ function explode_keycloak_deployments() {
...
+    fi
   done
 }
 
 function get_token() {
 
   token=""
   if [ -n "$SSO_USERNAME" ] && [ -n "$SSO_PASSWORD" ]; then
-    token=`$CURL --data "username=${SSO_USERNAME}&password=${SSO_PASSWORD}&grant_type=password&client_id=admin-cli" ${sso_service}/realms/${SSO_REALM}/protocol/openid-connect/token`
+    token=`$CURL --data "username=${SSO_USERNAME}&password=${SSO_PASSWORD}&grant_type=password&client_id=admin-cli" ${sso_service}/realms/master/protocol/openid-connect/token`
```
rcernich: Why did this change?  This would mean that all clients would need access to the master realm, even if they belonged to another realm.

rcernich: If the issue is that this doesn't work if `SSO_REALM` is not set, then it might be easier to set `SSO_REALM=${SSO_REALM:-master}` before configuration is started.

ruromero: @rcernich The admin-cli client_id only exists in the master realm. This doesn't mean that the client will be created in the master realm, this is for retrieving the token that will be use later to create the client.

rcernich: Right, but shouldn't the uid/pwd use correspond to an admin user in the specific realm?  If that's not the case, then I'm wondering if anybody has used this with a realm other than master.

ruromero: @rcernich apparently the only requirement is that the user $SSO_USERNAME exists on this realm to get the token. This token is only to create new clients which should be allowed to any user of the realm. I will do the change. 
Thanks for the comments.

ruromero:
@rcernich changes reverted

ruromero:
@rcernich can you review the changes? We need this to move forward, thanks

os-eap-sso/added/keycloak.sh
```diff
@@ -85,22 +75,22 @@ function configure_keycloak() {
     if [ -n "$SSO_SERVICE_URL" ]; then
       sso_service="$SSO_SERVICE_URL"
     fi
-  
+
+    if [ ! -n "${SSO_REALM:-master}" ]; then
```
rcernich: This will never be true.  Also, this either needs to be everywhere, or, after we test `-n "$SSO_REALM"`, we log the warning and `SSO_REALM=master` so the rest of the logic works correctly.

ruromero: @rcernich done, thanks.

spolti:
Does this PR fix this issue https://issues.jboss.org/browse/RHBA-646 ?

os-eap-sso/added/keycloak.sh
```diff
@@ -85,22 +75,22 @@ function configure_keycloak() {
     if [ -n "$SSO_SERVICE_URL" ]; then
       sso_service="$SSO_SERVICE_URL"
     fi
-  
+
+    if [ ! -n "${SSO_REALM}" ]; then
+      log_warning "Missing SSO_REALM. Defaulting to ${SSO_REALM:-master} realm"
```
rcernich: I think you actually need `SSO_REALM=master` here.  `${SSO_REALM:-master}` doesn't do anything to the value of `SSO_REALM`, it just evaluates to `master` if it is unset or empty.  You should be able to test this by passing an empty value in for `SSO_REALM`, which should yield a broken configuration.

ruromero: @rcernich you're right, I meant to use `:=` instead which sets the variable if empty

ruromero:
@rcernich can this be merged?

PR #220, opened 2018-03-27
CLOUD-2151 - Scaling pod down causes ActiveMQ AMQ214016 NoRouteToHost error messages

luck3y:
https://issues.jboss.org/browse/CLOUD-2151

Note this *removes* the embedded broker clustering, and thereby fixes the reported issue. The embedded broker clustering is not and never was officially supported.

Signed-off-by: Ken Wills <kwills@redhat.com

### DIFF
--- a/os-eap7-launch/added/launch/activemq-subsystem.xml
+++ b/os-eap7-launch/added/launch/activemq-subsystem.xml
@@ -1,5 +1,4 @@
     <server name="default">
-        <cluster password="${jboss.messaging.cluster.password:CHANGE ME!!}"/>
         <security-setting name="#">
             <role name="guest" send="true" consume="true" create-non-durable-queue="true" delete-non-durable-queue="true"/>
         </security-setting>
@@ -15,9 +14,6 @@
             <param name="direct-deliver" value="false"/>
         </http-acceptor>
         <in-vm-acceptor name="in-vm" server-id="0"/>
-        <broadcast-group name="bg-group1" connectors="http-connector" jgroups-channel="ee"/>
-        <discovery-group name="dg-group1" jgroups-channel="ee"/>
-        <cluster-connection name="my-cluster" address="jms" connector-name="http-connector" discovery-group="dg-group1"/>
         <jms-queue name="ExpiryQueue" entries="java:/jms/queue/ExpiryQueue"/>
         <jms-queue name="DLQ" entries="java:/jms/queue/DLQ"/>
         <!-- ##DESTINATIONS## -->
--- a/os-eap7-launch/added/launch/activemq-subsystem.xml
+++ b/os-eap7-launch/added/launch/activemq-subsystem.xml
@@ -18,6 +18,6 @@
         <jms-queue name="DLQ" entries="java:/jms/queue/DLQ"/>
         <!-- ##DESTINATIONS## -->
         <connection-factory name="InVmConnectionFactory" connectors="in-vm" entries="java:/ConnectionFactory"/>
-        <connection-factory name="RemoteConnectionFactory" ha="true" block-on-acknowledge="true" reconnect-attempts="-1" connectors="http-connector" entries="java:jboss/exported/jms/RemoteConnectionFactory"/>
+        <connection-factory name="RemoteConnectionFactory" connectors="http-connector" entries="java:jboss/exported/jms/RemoteConnectionFactory" reconnect-attempts="-1" />
         <pooled-connection-factory name="activemq-ra" transaction="xa" connectors="in-vm" entries="java:/JmsXA java:jboss/DefaultJMSConnectionFactory"/>
     </server>


luck3y:
Tested on EAP71 & EAPCD. The scale down exceptions are gone.

luck3y:
@rcernich & @bstansberry Review please.

luck3y:
Oh, good point, we should just like CLOUD-2404 to this, since it should take care of both. @bstansberry WDYT?

bstansberry:
@jmesnil Please approve.

@rcernich @luck3y If this goes in I think we can just close CLOUD-2404 as out of date, resolved due to fixing CLOUD-2151.

luck3y:
@rcernich Err, did I not include it? Sorry, I'll fix.

luck3y:
@rcernich corrected, thanks.

os-eap7-launch/added/launch/activemq-subsystem.xml
```diff
@@ -15,9 +14,6 @@
             <param name="direct-deliver" value="false"/>
         </http-acceptor>
         <in-vm-acceptor name="in-vm" server-id="0"/>
-        <broadcast-group name="bg-group1" connectors="http-connector" jgroups-channel="ee"/>
```
bstansberry: @jmesnil -- This is just me thinking out loud for the record.

If we leave this line inI believe this leaves the broker as discoverable by remote clients (say in an app w/ 1 messaging server pod and n > 1 front end pods that communicate with the messaging server.

BUT, leaving this in place also means if people deploy the front end onto the messaging server and scale that up, which seems a more likely scenario, then we get the problem of clients sending messages to random unrelated servers.

So, we shouldn't leave it. And that's consistent with the currently approved doc verbiage "ActiveMQ Artemis is only supported for messaging within a single pod with JBoss EAP instances, and is only enabled when JBoss AMQ is absent."

jmesnil: @bstansberry you are correct. broadcast-group (and discovery-group and cluster-connections) must be removed to ensure that this node will not advertise itself or connect to other nodes in any fashion.

This will isolate it from any other messaging servers

jmesnil: Removing HA support for the messaging subsystem should also require changes to the `RemoteConnectionFactory` so that is is configured as:

```
<connection-factory name="RemoteConnectionFactory" connectors="http-connector" entries="java:jboss/exported/jms/RemoteConnectionFactory" reconnect-attempsts="-1" />
```

This uses the default value for the attributes:
* `ha` (which is `false` as we have no other servers to failover)
* `block-on-acknowledge` (which is `false`). Blocking on acknowledgement is required or transparent failover. Since there are no other servers to failover, there is no good reason to keep blocking.

luck3y: Thanks Jeff, I'll check and update this too.

os-eap7-launch/added/launch/activemq-subsystem.xml
```diff
@@ -15,13 +14,10 @@
...
-        <discovery-group name="dg-group1" jgroups-channel="ee"/>
-        <cluster-connection name="my-cluster" address="jms" connector-name="http-connector" discovery-group="dg-group1"/>
         <jms-queue name="ExpiryQueue" entries="java:/jms/queue/ExpiryQueue"/>
         <jms-queue name="DLQ" entries="java:/jms/queue/DLQ"/>
         <!-- ##DESTINATIONS## -->
         <connection-factory name="InVmConnectionFactory" connectors="in-vm" entries="java:/ConnectionFactory"/>
-        <connection-factory name="RemoteConnectionFactory" ha="true" block-on-acknowledge="true" reconnect-attempts="-1" connectors="http-connector" entries="java:jboss/exported/jms/RemoteConnectionFactory"/>
+        <connection-factory name="RemoteConnectionFactory" connectors="http-connector" entries="java:jboss/exported/jms/RemoteConnectionFactory" reconnect-attempts="-1" />
```
maschmid: Should it still be reconnect-attempts="-1" for non-ha?  (as this option isn't being set in non-ha version of standalone-full.xml )

luck3y: @jmesnil Is the non-HA default of econnect-attempts -1?

jmesnil: yes, keeping -1 (infinite reconnection) is correct.
By default, in non-HA case, Artemis clients never tries to reconnect (reconnect-attempts = 0).
In cloud environment, it is expected that the Artemis servers should always be up.
It might not be the case during deployment rolling upgrades of the Artemis servers but keeping reconnect-attempts to -1 ensures that the Artemis clients will "wait" until the Artemis servers are up again.

PR #226, opened 2018-04-10
[CLOUD-2455] - [BxMS] ERROR IJ000315: Pool amq-ConnectionFactory has …

spolti:
…1 active handles

Signed-off-by: Filippe Spolti <fspolti@redhat.com>

### DIFF
--- a/os-eap-launch/added/launch/resource-adapters-common.sh
+++ b/os-eap-launch/added/launch/resource-adapters-common.sh
@@ -18,6 +18,7 @@ function clearResourceAdapterEnv() {
   unset ${prefix}_RECOVERY_USERNAME
   unset ${prefix}_RECOVERY_PASSWORD
   unset ${prefix}_ADMIN_OBJECTS
+  unset ${prefix}_TRACKING
 
   for xa_prop in $(compgen -v | grep -s "${prefix}_PROPERTY_"); do
     unset ${xa_prop}
@@ -106,7 +107,14 @@ function inject_resource_adapters_common() {
       resource_adapter="${resource_adapter}<transaction-support>$transaction_support</transaction-support>"
     fi
 
-    resource_adapter="${resource_adapter}<connection-definitions><connection-definition class-name=\"${ra_class}\" jndi-name=\"${ra_jndi}\" enabled=\"true\" use-java-context=\"true\">"
+    resource_adapter="${resource_adapter}<connection-definitions><connection-definition"
+
+    tracking=$(find_env "${ra_prefix}_TRACKING")
+    if [ -n "${tracking}" ]; then
+      # monitor applications, look for unclosed resources.
+      resource_adapter="${resource_adapter} tracking=\"${tracking}\""
+    fi
+    resource_adapter="${resource_adapter} class-name=\"${ra_class}\" jndi-name=\"${ra_jndi}\" enabled=\"true\" use-java-context=\"true\">"
 
     ra_props=$(compgen -v | grep -s "${ra_prefix}_PROPERTY_")
     if [ -n "$ra_props" ]; then
--- a/os-eap7-launch/added/launch/messaging.sh
+++ b/os-eap7-launch/added/launch/messaging.sh
@@ -156,6 +156,7 @@ function generate_object_config() {
 # $10 - driver
 # $11 - queue names
 # $12 - topic names
+# $13 - ra tracking
 function generate_resource_adapter() {
   log_info "Generating resource adapter configuration for service: $1 (${10})" >&2
   IFS=',' read -a queues <<< ${11}
@@ -172,6 +173,7 @@ function generate_resource_adapter() {
                     <config-property name=\"ServerUrl\">tcp://$6:$7?jms.rmIdFromConnectionId=true</config-property>
                     <connection-definitions>
                         <connection-definition
+                              "${13}"
                               class-name=\"org.apache.activemq.ra.ActiveMQManagedConnectionFactory\"
                               jndi-name=\"$2\"
                               enabled=\"true\"
@@ -313,6 +315,11 @@ function inject_brokers() {
       # topics environment variable name format: [NAME]_[BROKER_TYPE]_TOPICS
       topics=$(find_env "${prefix}_TOPICS")
 
+      tracking=$(find_env "${prefix}_TRACKING")
+      if [ -n "${tracking}" ]; then
+         ra_tracking="tracking=\"${tracking}\""
+      fi
+
       case "$type" in
         "AMQ")
           driver="amq"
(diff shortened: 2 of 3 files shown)


os-eap-launch/added/launch/resource-adapters-common.sh
```diff
@@ -106,7 +107,13 @@ function inject_resource_adapters_common() {
       resource_adapter="${resource_adapter}<transaction-support>$transaction_support</transaction-support>"
     fi
 
-    resource_adapter="${resource_adapter}<connection-definitions><connection-definition class-name=\"${ra_class}\" jndi-name=\"${ra_jndi}\" enabled=\"true\" use-java-context=\"true\">"
+    resource_adapter="${resource_adapter}<connection-definitions><connection-definition"
+    # monitor applications, look for unclosed resources.
+    tracking=$(find_env "${ra_prefix}_TRACKING" "true")
```
errantepiphany: 1. I think this variable can be declared as local, and moved into the "if [[ "$JBOSS_EAP_VERSION" == "7."* ]]" statement below it.
2. Shouldn't the default be "false", per the workaround described in CLOUD-2455?

spolti: agreed, thanks.

os-eap-launch/added/launch/resource-adapters-common.sh
```diff
@@ -106,7 +107,14 @@ function inject_resource_adapters_common() {
       resource_adapter="${resource_adapter}<transaction-support>$transaction_support</transaction-support>"
     fi
 
-    resource_adapter="${resource_adapter}<connection-definitions><connection-definition class-name=\"${ra_class}\" jndi-name=\"${ra_jndi}\" enabled=\"true\" use-java-context=\"true\">"
+    resource_adapter="${resource_adapter}<connection-definitions><connection-definition"
+
+    if [[ "$JBOSS_EAP_VERSION" == "7."* ]]; then
+      # monitor applications, look for unclosed resources.
+      local tracking=$(find_env "${ra_prefix}_TRACKING" "false")
```
maschmid: we should keep the default and set it only if the property is set.   (I believe in general it default to true, but perhaps not always)

maschmid:
Make sure you also don't need to change 

https://github.com/jboss-openshift/cct_module/blob/master/os-eap7-launch/added/launch/messaging.sh#L174

(or do that same / similar change there if you need)

os-eap-launch/added/launch/resource-adapters-common.sh
errantepiphany: After discussion, I agree with @maschmid . Instead of checking the jboss eap version, just see if the tracking variable is set. If it is, set it, otherwise, do nothing. Additionally, instead of having to change all our KIE templates to set this variable to false, we can just set them in the image.yaml (not even needing to declare them in the templates).

spolti: The problem of doing this is that it will also be applied on eap6, is this setting also necessary on eap6?

errantepiphany: The idea is that we wouldn't set the variable unless the en var is defined (so, don't use a default value when looking up the env var). It then wouldn't be applied to eap6, because no one would be setting the env var.

os-eap-launch/added/launch/resource-adapters-common.sh
```diff
@@ -106,7 +107,13 @@ function inject_resource_adapters_common() {
       resource_adapter="${resource_adapter}<transaction-support>$transaction_support</transaction-support>"
     fi
 
-    resource_adapter="${resource_adapter}<connection-definitions><connection-definition class-name=\"${ra_class}\" jndi-name=\"${ra_jndi}\" enabled=\"true\" use-java-context=\"true\">"
+    resource_adapter="${resource_adapter}<connection-definitions><connection-definition"
+
+    if [ -n "${RA_TRACKING}" ]; then
```
maschmid: in this case this is resource-adapter specific, so it should use the ra_prefix_TRACKING

os-eap7-launch/added/launch/messaging.sh
```diff
@@ -313,14 +315,18 @@ function inject_brokers() {
       # topics environment variable name format: [NAME]_[BROKER_TYPE]_TOPICS
       topics=$(find_env "${prefix}_TOPICS")
 
+      if [ -n "${RA_TRACKING}" ]; then
```
maschmid: should be specific to the resource adapter, as in general there can be more  (should use the ${prefix}_TRACKING

os-eap7-launch/added/launch/messaging.sh
```diff
@@ -172,6 +173,7 @@ function generate_resource_adapter() {
                     <config-property name=\"ServerUrl\">tcp://$6:$7?jms.rmIdFromConnectionId=true</config-property>
                     <connection-definitions>
                         <connection-definition
+                              "${13}"
```
maschmid: this is already inside quotes , did you mean just ${13} ?

os-eap7-launch/added/launch/messaging.sh
```diff
@@ -313,14 +315,19 @@ function inject_brokers() {
       # topics environment variable name format: [NAME]_[BROKER_TYPE]_TOPICS
       topics=$(find_env "${prefix}_TOPICS")
 
+      tracking=$(find_env "${prefix}_TRACKING")
+      if [ -n "${tracking}" ]; then
+         ra_tracking="tracking=\"${tracking}\""
+      fi
```
maschmid: ra_tracking is pontentially undefined here, it would be better to explicitly set it to an empty string, just to be safe from a global "ra_tracking" set somewhere else...

os-eap7-launch/added/launch/messaging.sh
spolti: Its ok use quotes there since it is not escaped.

os-eap7-launch/added/launch/messaging.sh
spolti: It is a hard scenario but can happen since this var is not local, I'll updated it, does it block the QE to test using the images that is already there?

maschmid: no I don't think we need to retest on trivial changes.

os-eap-launch/added/launch/resource-adapters-common.sh
```diff
@@ -106,7 +107,14 @@ function inject_resource_adapters_common() {
       resource_adapter="${resource_adapter}<transaction-support>$transaction_support</transaction-support>"
     fi
 
-    resource_adapter="${resource_adapter}<connection-definitions><connection-definition class-name=\"${ra_class}\" jndi-name=\"${ra_jndi}\" enabled=\"true\" use-java-context=\"true\">"
+    resource_adapter="${resource_adapter}<connection-definitions><connection-definition"
+
+    tracking=$(find_env "${ra_prefix}_TRACKING")
```
rcernich: This should be local.

rcernich: never mind.  it looks like this is initialized every time.

PR #228, opened 2018-04-12
[CLOUD-2465] Allow custom location for local maven repository

bmozaffa:
https://issues.jboss.org/browse/CLOUD-2465
Allow custom location for local maven repository by using <localRepository> to designate a location other than the standard ~/.m2/repository

### DIFF
--- a/os-eap-launch/added/launch/maven-repos.sh
+++ b/os-eap-launch/added/launch/maven-repos.sh
@@ -7,6 +7,7 @@ function prepareEnv() {
     unset MAVEN_REPO_HOST
     unset MAVEN_REPO_ID
     unset MAVEN_REPO_LAYOUT
+    unset MAVEN_REPO_LOCAL
     unset MAVEN_REPO_PASSPHRASE
     unset MAVEN_REPO_PASSWORD
     unset MAVEN_REPO_PATH
@@ -32,6 +33,10 @@ function configure() {
 
 function configure_maven_repos {
     local settings="${1-$HOME/.m2/settings.xml}"
+    local local_repo_path="${MAVEN_REPO_LOCAL}"
+    if [ "${local_repo_path}" != "" ]; then
+        set_local_repo_path "${settings}" "${local_repo_path}"
+    fi
     # single repo scenario: respect fully qualified url if specified, otherwise find and use service
     local single_repo_url="${MAVEN_REPO_URL}"
     if [ "${single_repo_url}" = "" ]; then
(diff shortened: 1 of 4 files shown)


luck3y:
Would this not be better done by adding a profile to the existing settings.xml in the product that needs it?

https://github.com/jboss-openshift/cct_module/blob/master/jboss-maven/added/jboss-settings.xml#L274

errantepiphany:
@luck3y Whereas I think that would work, it seems like overkill for an otherwise very simple config point.

luck3y:
@errantepiphany Not sure about that. If you want to use profiles from both this maven config and the default one, you'd have to duplicate the old ones into the new config.

luck3y:
Yeah, never mind I thought this was replacing the existing one with the new one instead adding to it. Seems OK to me.

bmozaffa:
Updated PR to include the secondary settings.xml file in s2i so that it too can take advantage of this parameter: https://github.com/bmozaffa/cct_module/blob/d861242758c906a47eb748e0edbdeebf441dfcca/os-java-s2i/added/s2i/jboss-settings.xml#L6

rcernich:
Yes, the jboss-settings.xml is shared, sort of.  The maven-repos.sh file is only used by kie images.  Common maveny stuff is in s2i-common.  Since the local repo location can be set using a system property, is it really necessary to add this into settings.xml?

bmozaffa:
@rcernich Replied on the jira too... I'd have to add it to our queue for someone to verify, but my concern is that JVM system properties may not be propagated to the maven process in business-central, deliberately, to ensure a segregated build environment.

errantepiphany:
@rcernich , please see my comment on the jira. The property approach does not work.

os-eap-launch/added/launch/maven-repos.sh
```diff
@@ -177,3 +182,12 @@ function generate_random_id() {
     cat /dev/urandom | env LC_CTYPE=C tr -dc 'a-zA-Z0-9' | fold -w 16 | head -n 1
 }
 
+function set_local_repo_path() {
+    local settings="${1}"
+    local local_path="${2}"
+    local xml="\n\
+    <localRepository>${local_path}</localRepository>\n\
+    <!-- ### configured local repository ### -->"
```
errantepiphany: We don't need to put the "configured local repository" comment line back in here, because multiple search and replaces are not supported. So I think Line 187 through 190 could just be replaced with:
local xml="<localRepository>${local_path}</localRepository>"

PR #229, opened 2018-04-13
[CLOUD-2068] - make refresh-interval for timer data-store configurable

spolti:
Signed-off-by: Filippe Spolti <fspolti@redhat.com>


os-eap-launch/added/launch/datasource-common.sh
```diff
@@ -197,7 +197,11 @@ function generate_datasource_common() {
 
   if [ -n "$TIMER_SERVICE_DATA_STORE" -a "$TIMER_SERVICE_DATA_STORE" = "${service_name}" ]; then
     inject_timer_service ${pool_name}_ds
-    inject_datastore $pool_name $jndi_name $driver
+    local refresh_interval
+    if [[ "$JBOSS_EAP_VERSION" == "7."* ]]; then
```
rcernich: Unfortunately, this will not work across all EAP based products as JBOSS_EAP_VERSION may not defined (as a matter of fact, it's only defined for EAP 6.4 and EAP 7.x images).  This will need to be broken out into the EAP 7 specific datasource configuration script.

spolti: Good catch Rob, thanks, I'll rework it and update this pr for your review.

errantepiphany: @spolti Status? This needs to get updated per @rcernich 's note above before RHPAM-181 can move forward. Thanks!

spolti: @errantepiphany I am testing this change with eap images and also with rhpam kieserver image.
Probably by tomorrow morning I have this pr updated.

errantepiphany:
@rcernich Are you okay now with @spolti's latest update?

os-eap64-launch/added/launch/datasource.sh
```diff
@@ -26,3 +26,8 @@ function inject_datasources() {
 function generate_datasource() {
   generate_datasource_common "${1}" "${2}" "${3}" "${4}" "${5}" "${6}" "${7}" "${8}" "${9}" "${10}" "${11}" "${12}" "${13}" "${14}"
 }
+
+function refresh_interval() {
```
rcernich: Would it be better to stub this out in datasource-common.sh, then have it be overridden by the datasource.sh file that is including it?  My concern is that this is a required method, but may not be available in other scripts that utilize it (i.e. we may break existing users who are not aware of this change).  A simple test would be to move this implementation into datasource-common.sh and use it as the default.

spolti:
@rcernich can you please take a look?

PR #248, opened 2018-05-14
CLOUD-2391 Fixed REST access logging

slaskawi:
https://issues.jboss.org/browse/CLOUD-2391


slaskawi:
@rcernich Right! Very good point. I totally forgot that we work a big differently than all other projects (most of them put code into the master and cut off stable branches out of it; we do the other way around). Will implement a fix in a moment.

Again, thanks for pointing it out.

slaskawi:
Closed in favor of https://github.com/jboss-openshift/cct_module/pull/252

PR #249, opened 2018-05-14
[CLOUD-2437] Introduce the ability to set JGroups protocol type to ASYM_ENCRYPT

iankko:
Perform this by introduction of the new ```JGROUPS_ENCRYPT_PROTOCOL``` environment variable, which can accept one of the two possible values:
    
  - Either ```SYM_ENCRYPT``` (default setting when ```JGROUPS_ENCRYPT_SECRET``` variable is set),
  - Or ```ASYM_ENCRYPT``` (after explicit request)

Also describe the new ```JGROUPS_ENCRYPT_PROTOCOL``` environment variable in the ```module.yaml``` definition of the ```os-eap-launch``` module

Signed-off-by: Jan Lieskovsky <jlieskov@redhat.com>


iankko:
@rcernich @maschmid PTAL (once got a chance)

Thank you, Jan

os-eap-launch/added/launch/jgroups.sh
```diff
@@ -36,6 +37,20 @@ function configure_jgroups_encryption() {
     else
       log_warning "Partial JGroups encryption configuration, the communication within the cluster WILL NOT be encrypted."
     fi
+  elif [ -n "${JGROUPS_ASYM_ENCRYPT}" ] && [ "${JGROUPS_ASYM_ENCRYPT,,}" == "true" ]; then
```
rcernich: Would it be better to have something like `JGROUPS_ENCRYPT_PROTOCOL` instead, e.g. `JGROUPS_ENCRYPT_PROTOCOL=ASYM_ENCRYPT`?  This would default to `SYM_ENCRYPT` if `JGROUPS_ENCRYPT_NAME` is specified, to preserve backward compatibility.  This would allow you to key of the desired protocol and do whatever is necessary for that protocol.

iankko: @rcernich Good point / suggestion.

Agree ```JGROUPS_ENCRYPT_PROTOCOL``` is better name. Will update yet (plus change the logic the ```SYM_ENCRYPT``` protocol to be the default one).

iankko:
@rcernich 
> Will these changes break existing apps that may be looking for jgroups keystore in /etc/x509/jgroups? Have you tested with an app that was instantiated using the old template/image, then updated it to this version of the image?

Performed the testing using ```oc new-app --template=sso72-x509-https``` command using the old template version and old image. Then edited ```dc/sso``` to add  ```JGROUPS_ASYM_ENCRYPT=true``` variable:

```
$ oc set env dc/sso -e JGROUPS_ASYM_ENCRYPT="true"
```
and rolled out the ```latest``` image, where ```latest``` has been modified to point to the updated image:

```
$ oc rollout latest dc/sso
deploymentconfig "sso" rolled out
```
and AFAICT this change doesn't seem to have a negative effect (nothing suspicious in the log of the pod).

Will perform the change to replace ```JGROUPS_ASYM_ENCRYPT``` with ```JGROUPS_ENCRYPT_PROTOCOL``` as suggested above yet.

iankko:
@rcernich Hey Rob, jFYI, the PR was updated to include the ```JGROUPS_ENCRYPT_PROTOCOL``` variable. Should be ready for re-review.

Thanks, Jan

os-sso72/added/launch/openshift-x509.sh
```diff
@@ -8,69 +8,59 @@ function prepareEnv() {
...
+function autogenerate_https_keystore_and_rhsso_truststore() {
   # Keystore infix notation as used in templates to keystore name mapping
-  declare -A KEYSTORES=( ["https"]="HTTPS" ["jgroups"]="JGroups" )
-
-  # CLOUD-2436 Don't generate the JGroups keystore if custom
-  # JGROUPS_ENCRYPT_SECRET was provided
-  if [ -n "${JGROUPS_ENCRYPT_SECRET}" ]; then
-    unset KEYSTORES["jgroups"]
-  fi
+  declare -A KEYSTORES=( ["https"]="HTTPS" )
```
Crumby: This array looks unnecessary...

iankko: @Crumby Great catch, Pavel! Dropped, thanks!

os-eap-launch/added/launch/jgroups.sh
```diff
@@ -22,21 +22,41 @@ function configure_jgroups_encryption() {
...
-          <property name=\"sym_algorithm\">AES</property>\
-          <property name=\"encrypt_entire_message\">true</property>\
-          <property name=\"keystore_name\">${JGROUPS_ENCRYPT_KEYSTORE_DIR}/${JGROUPS_ENCRYPT_KEYSTORE}</property>\
-          <property name=\"store_password\">${JGROUPS_ENCRYPT_PASSWORD}</property>\
-          <property name=\"alias\">${JGROUPS_ENCRYPT_NAME}</property>\
-        </protocol>"
-    else
-      log_warning "Partial JGroups encryption configuration, the communication within the cluster WILL NOT be encrypted."
-    fi
+    JGROUPS_ENCRYPT_PROTOCOL="SYM_ENCRYPT"
```
rcernich: Won't this always produce `SYM_ENCRYPT`?  Shouldn't it be `JGROUPS_ENCRYPT_PROTOCOL="${JGROUPS_ENCRYPT_PROTOCOL:-SYM_ENCRYPT}"`?

iankko: Correct. It will always produce ```SYM_ENCRYPT``` when the ```JGROUPS_ENCRYPT_SECRET``` is also set. The JGroups ASYM_ENCRYPT protocol will be set if and only if:

- ```JGROUPS_ENCRYPT_PROTOCOL``` is set to ```ASYM_ENCRYPT```, and
- None of ```JGROUPS_ENCRYPT_SECRET```, ```JGROUPS_ENCRYPT_NAME```, and ```JGROUPS_ENCRYPT_PASSWORD``` are set. Per JGroups documentation, ASYM_ENCRYPT doesn't need these (the JGroups keystore definition), since it will generate the public / private key pair by itself under the hood.

So the assumption was, if the user provided JGroups keystore, they want ```SYM_ENCRYPT```. When they don't provide it, they want ```ASYM_ENCRYPT```.

rcernich: I think we default it if it's not set.  If they specify both, we just log a warning, e.g. "JGROUPS_ENCRYPT_SECRET ignored when configuring ASYM_ENCRYPT."  This provides a more natural use case, where folks might want to migrate to ASYM_ENCRYPT, but forget to nuke the other variable setting.

iankko: Fair enough. Your proposal looks reasonable. Will update the PR yet. Thanks for these suggestions!

iankko:
@rcernich 

Hey Rob, jFYI, this PR has been updated to address your suggestions WRT to the new ```JGROUPS_ENCRYPT_PROTOCOL``` environment variable.

Also, due to the backwards compatibility, checked the suggested scenario:
> Will these changes break existing apps that may be looking for jgroups keystore in /etc/x509/jgroups? Have you tested with an app that was instantiated using the old template/image, then updated it to this version of the image?

and
> Regarding backward compatibility, what happens if the user doesn't oc set env dc/sso -e JGROUPS_ASYM_ENCRYPT="true"? That's the user migrating their deployment to work with the new image.

Please see below:
- Deployed both the old (unmodified) image and old template,
- Then rolled out the updated image (using the tag, the old template expects). Just warning about partial JGroups configuration is printed:

jgroups_warning

So IMHO, this is OK, correct?

Thanks, Jan

rcernich:
I guess that depends: what was jgroups doing before?  If was encrypted and it's still encrypted in the same way, then I would consider that backward compatible.  If it was encrypted and it's no longer encrypted or it's encrypted differently, then that would be incompatible.  I.e., is the user still getting what they asked for originally?

os-eap-launch/added/launch/jgroups.sh
```diff
@@ -21,22 +21,41 @@ function configureEnv() {
...
+        jgroups_encrypt="<protocol type=\"SYM_ENCRYPT\">\n\
+                        <property name=\"provider\">SunJCE</property>\n\
+                        <property name=\"sym_algorithm\">AES</property>\n\
+                        <property name=\"encrypt_entire_message\">true</property>\n\
+                        <property name=\"keystore_name\">${JGROUPS_ENCRYPT_KEYSTORE_DIR}/${JGROUPS_ENCRYPT_KEYSTORE}</property>\n\
+                        <property name=\"store_password\">${JGROUPS_ENCRYPT_PASSWORD}</property>\n\
+                        <property name=\"alias\">${JGROUPS_ENCRYPT_NAME}</property>\n\
+                    </protocol>"
+      else
+        log_warning "Partial JGroups encryption configuration, the communication within the cluster WILL NOT be encrypted."
```
Crumby: This warning will be always generated if user didn't set any environment regarding jgroups. Previous version printed warning only if SECRET envs was set and either USERNAMe or PASSWORD was missing. Is this what we want?

iankko: @Crumby 
Good catch, Pavel! The image has been updated (see most recent version) that warning message to distinguish between the cases the JGroups configuration is partial, or not set at all (the cluster traffic wouldn't be encrypted in none of these cases), and based on that print the appropriate state.

iankko:
@Crumby Updated the PR per your feedback.

The only difference between the current and proposed images (besides the introduction of ASYM_ENCRYPT feature of course) is, the updated image will print warning like depicted on the following image:

warning

in the case ```SYM_ENCRYPT``` (the default) cluster traffic encryption method was chosen, but ```JGROUPS_ENCRYPT_SECRET``` wasn't simultaneously defined (the current image prints the warning only in case the JGroups encryption config is partial, but not completely missing).

But IMHO the warning about the missing JGroups encryption config is adequate (the user should be aware, the SYM_ENCRYPT setting won't have an effect, if they don't configure JGROUPS_ENCRYPT_SECRET variable too). But I can completely remove it (the image behaviour when using SYM_ENCRYPT would be identical between current and proposed images), if you think that message is inappropriate / want it to be removed.

Let me know.

Thank you, Jan

os-eap-launch/added/launch/jgroups.sh
```diff
@@ -21,22 +21,51 @@ function configureEnv() {
...
-    fi
-  fi
+  JGROUPS_ENCRYPT_PROTOCOL="${JGROUPS_ENCRYPT_PROTOCOL:=SYM_ENCRYPT}"
+
+  case "${JGROUPS_ENCRYPT_PROTOCOL}" in
+    "SYM_ENCRYPT")
+      log_info "Configuring JGroups cluster traffic encryption protocol to SYM_ENCRYPT."
+      local JGROUPS_UNENCRYPTED_MESSAGE="Detected <STATE> JGroups encryption configuration, the communication within the cluster WILL NOT be encrypted."
+      local KEYSTORE_ERROR_MESSAGE=""
+      [ -n "${JGROUPS_ENCRYPT_SECRET}" -a -n "${JGROUPS_ENCRYPT_NAME}" -a -n "${JGROUPS_ENCRYPT_PASSWORD}" ]
```
Crumby: @iankko Why do we have JGROUPS_ENCRYPT_SECRET env var? Does it something somewhere? In this script it's only check whether it exists and it has to exist, but no useful information is read from it. Also what if user doesn't specify JGROUPS_ENCRYPT_KEYSTORE_DIR and JGROUPS_ENCRYPT_KEYSTORE which are used for configuration?

luck3y: @Crumby JGROUPS_ENCRYPT_SECRET is a reference to the secret installed by the app, template example here: https://github.com/jboss-container-images/jboss-eap-7-openshift-image/blob/eap-cd-dev/templates/eap-cd-sso-s2i.json#L828

I think the secret is a prerequisite for having encryption in jgroups.

iankko: @Crumby That's actually a very good question (re: why do we have JGROUPS_ENCRYPT_SECRET). I would assume this is needed in the templates to be able to mount the [`eap-jgroups-keystore-volume`](https://github.com/jboss-openshift/application-templates/blob/master/eap/eap71-https-s2i.json#L608). But you are correct, even if supplied it's not used for actual setting of JGroups protocol type.

@rcernich -^ Can you clarify why it was / is still needed? IOW is the only usage of the `JGROUPS_ENCRYPT_SECRET` variable to specify the secret, to be used as the source for the [volume mount](https://docs.openshift.com/container-platform/3.10/dev_guide/volumes.html#adding-volumes) (see `-t` option in `Table 2. Supported options for Adding Volumes` in the previous link). Would it be possible to use configmap in the same way? If so, should we enhance the support?

@Crumby For your second question (re: `Also what if user doesn't specify JGROUPS_ENCRYPT_KEYSTORE_DIR and JGROUPS_ENCRYPT_KEYSTORE which are used for configuration?`)

From looking at the template, the `JGROUPS_ENCRYPT_KEYSTORE_DIR` is set [always](https://github.com/jboss-openshift/application-templates/blob/master/eap/eap71-https-s2i.json#L574), when using the template. But `JGROUPS_ENCRYPT_KEYSTORE` might be undefined (isn't required param AFAICT). In that case the volume mount would fail, and the pod would get stuck in 'Pending' state. For deploying the image directly, both of them need to be specified (if not, again the pod would be stuck keep trying to mount the volume).

iankko:
Compared with the previous version, the current form of this PR was equipped with two additional commits:

```
commit 9c58068437ef1263e21b26b673b664c31fdbf3bf
Author: Jan Lieskovsky <jlieskov@redhat.com>
Date:   Fri Aug 10 15:36:45 2018 +0200

    [CLOUD-2437] Require AUTH protocol (JGROUPS_CLUSTER_PASSWORD environment variable)
    to be set when using the JGroups ASYM_ENCRYPT cluster traffic encryption protocol
    
    Signed-off-by: Jan Lieskovsky <jlieskov@redhat.com>
```
and
```
commit 04e4a542819f57c89818c990c0ab49e83ef391ef (HEAD -> CLOUD-2437, origin/CLOUD-2437)
Author: Jan Lieskovsky <jlieskov@redhat.com>
Date:   Fri Aug 10 17:31:20 2018 +0200

    [CLOUD-2437] Consolidate the JGroups CTF tests for various images under eap_common.feature
    
    This commit does the following:
    * Put the (outdated) copies of datavirt/jgroups_encrypt.feature and
      sso/jgroups_encrypt.feature CTF tests under one umbrella located
      in eap/eap_common.feature,
    * Update the existing SYM_ENCRYPT JGroups CTF tests to start using
      the JGROUPS_ENCRYPT_PROTOCOL environment variable
    * Introduce ASYM_ENCRYPT JGroups encryption protocol CTF tests:
      - Check if JGROUPS_CLUSTER_PASSWORD is required when using ASYM_ENCRYPT
      - Check if warning is issues, when using ASYM_ENCRYPT and some of
        JGROUPS_ENCRYPT_SECRET, JGROUPS_ENCRYPT_NAME, JGROUPS_ENCRYPT_PASSWORD,
        JGROUPS_ENCRYPT_KEYSTORE_DIR, or JGROUPS_ENCRYPT_KEYSTORE variables
        is set
    
    Signed-off-by: Jan Lieskovsky <jlieskov@redhat.com>
```

Also the check for (possibly missing) `JGROUPS_ENCRYPT_KEYSTORE_DIR` and `JGROUPS_ENCRYPT_KEYSTORE` variables, when using `SYM_ENCRYPT` protocol was squashed into  30480105729e6f9ba11f0ff14e52a4183751d72d commit.

@Crumby @rcernich @luck3y Should be ready for re-review.

Thank you, Jan

os-eap64-launch/added/launch/ha.sh
```diff
@@ -72,9 +72,22 @@ function configure_ha() {
   JBOSS_HA_ARGS="${JBOSS_HA_ARGS} -Djboss.node.name=${JBOSS_NODE_NAME}"
 
   if [ -z "${JGROUPS_CLUSTER_PASSWORD}" ]; then
-      log_warning "No password defined for JGroups cluster. AUTH protocol will be disabled. Please define JGROUPS_CLUSTER_PASSWORD."
-      JGROUPS_AUTH="<!--WARNING: No password defined for JGroups cluster. AUTH protocol has been disabled. Please define JGROUPS_CLUSTER_PASSWORD. -->"
+    local NL=$'\n'
+    # CLOUD-2437 AUTH protocol is required when using ASYM_ENCRYPT protocol: https://github.com/belaban/JGroups/blob/master/conf/asym-encrypt.xml#L23
+    # And is optional when using SYM_ENCRYPT protocol. Adjust the AUTH_PROTO_WARNING_MESSAGE below depending on JGroups encryption protocol being used
+    local AUTH_PROTO_WARNING_MESSAGE="No password defined for JGroups cluster. AUTH protocol <AUTH_STATE> Please define JGROUPS_CLUSTER_PASSWORD."
+    if [ "${JGROUPS_ENCRYPT_PROTOCOL}" == "ASYM_ENCRYPT" ]; then
```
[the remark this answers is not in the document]
Crumby: Also, take a look at the log order with this change:
```

WARN No password defined for JGroups cluster. AUTH protocol is required when using JGroups ASYM_ENCRYPT cluster traffic encryption protocol.           
The communication within the cluster WILL NOT be encrypted. Please define JGROUPS_CLUSTER_PASSWORD.
INFO Service account has sufficient permissions to view pods in kubernetes (HTTP 200). Clustering will be available.
sed: -e expression #1, char 188: unterminated `s' command
INFO Configuring JGroups discovery protocol to openshift.KUBE_PING
INFO Configuring JGroups cluster traffic encryption protocol to ASYM_ENCRYPT.
```

First it outputs a warning that no configuration will happen and then it outputs it's configuring....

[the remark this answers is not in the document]
iankko: @Crumby Thanks, Pavel. Yeah, that unterminated warning is syntax error in sed command used. Will have a look yet.

iankko:
Replaced by https://github.com/jboss-openshift/cct_module/pull/288 one.

PR #252, opened 2018-05-16
CLOUD-2391 Fixed REST access logging

slaskawi:
https://issues.jboss.org/browse/CLOUD-2391


slaskawi:
@rcernich in https://github.com/jboss-openshift/cct_module/pull/248 you mentioned:

> you might want to move your work to a working branch to prevent merge issues in your master branch, e.g. if this commit is rebased before merging upstream.

I moved it here (to branch `sprint-16`) but according to the latest email, we should be working on `sprint-18`, shouldn't we? But it seems that branch is not there....

PR #253, opened 2018-05-16
CLOUD-2225 - [EAP] cannot define multiple AMQ resource adapters

luck3y:
https://issues.jboss.org/browse/CLOUD-2225

This turned out a bit tricky to test. There are a couple of scenarios:

(1) EAP & AMQ, using the template, one RA, should work as before.
(2) EAP using two RAs defined like:

Two AMQ brokers already setup, with name like below, then adding EAP:

oc new-app eap71-basic-s2i \
-p APPLICATION_NAME=eap71-$1 \
-e MQ_SERVICE_PREFIX_MAPPING="broker1-amq=BROKER1_MQ broker2-amq=BROKER2_MQ" \
-e BROKER1_AMQ_TCP_SERVICE_HOST=broker1-amq-tcp \
-e BROKER1_AMQ_TCP_SERVICE_PORT=61616 \
-e BROKER1_MQ_USERNAME=test1user \
-e BROKER1_MQ_PASSWORD=test1pass \
-e BROKER2_AMQ_TCP_SERVICE_HOST=broker2-amq-tcp \
-e BROKER2_AMQ_TCP_SERVICE_PORT=61616 \
-e BROKER2_MQ_USERNAME=test2user \
-e BROKER2_MQ_PASSWORD=test2pass 
 
These are now named activemq-ra.rar and then activemq-ra.rar-1, -2, -3 etc.

An additional variable EJB_RESOURCE_ADAPTER_NAME is introduced and if not set, is defaulted to the first RA configured, to try and preserve compatabilty with the old behavior.

### DIFF
--- a/os-eap7-launch/added/launch/messaging.sh
+++ b/os-eap7-launch/added/launch/messaging.sh
@@ -157,15 +157,30 @@ function generate_object_config() {
 # $11 - queue names
 # $12 - topic names
 # $13 - ra tracking
+# $14 - resource counter, incremented for each broker, starting at 0
 function generate_resource_adapter() {
   log_info "Generating resource adapter configuration for service: $1 (${10})" >&2
   IFS=',' read -a queues <<< ${11}
   IFS=',' read -a topics <<< ${12}
+
+  local ra_id=""
+  # this preserves the expected behavior of the first RA, and doesn't append a number. Any additional RAs will have -count appended.
+  if [ "${14}" -eq "0" ]; then
+    ra_id="${9}"
+  else
+    ra_id="${9}-${14}"
+  fi
+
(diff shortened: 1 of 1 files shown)


luck3y:
@rcernich I started out using name-number for all of them, but then I worried maybe people had legacy deployment configs and activation props that referenced the legacy name, so thought that it might be a good idea to try to remain compatible.

WRT to naming, it can be used in some activation props, but since they couldn't change the name before, I'm not sure it would have been useful. If we do want to do that, I'd suggest we do it as a different enhancement though, as it'd be allowing them to override, since we'll still need reasonable defaults.

rcernich:
Cool.  That works for me.  Good job.

luck3y:
Closing this for now, will re-open against jboss-eap-modules.

PR #256, opened 2018-06-01
Copy os-sso72 to os-sso73

ASzc:


ASzc:
Needed for [here](https://github.com/jboss-container-images/redhat-sso-7-openshift-image/blob/sso73-dev/image.yaml#L82). Copying this for each version doesn't seem the best that we could be doing, but if we don't have any other system ready, then this will have to do.

PR #268, opened 2018-07-16
[CLOUD-2286] permit user configuration of GC

jmtd:
Move hard-coded GC options into a "GC_OPTIONS" environment variable
that can be user-supplied to override the defaults.

https://issues.jboss.org/browse/CLOUD-2286

### DIFF
--- a/os-java-run/module.yaml
+++ b/os-java-run/module.yaml
@@ -38,3 +38,6 @@ envs:
     - name: GC_MAX_METASPACE_SIZE
       description: The maximum metaspace size.
       example: "100"
+    - name: GC_CONTAINER_OPTIONS
+      description: specify Java GC to use. The value of this variable should contain the necessary JRE command-line options to specify the required GC, which will override the default of `-XX:+UseParallelOldGC`.
+      example: -XX:+UseG1GC
--- a/os-java-run/added/java-default-options
+++ b/os-java-run/added/java-default-options
@@ -117,10 +117,11 @@ gc_config() {
   local timeRatio=${GC_TIME_RATIO:-4}
   local adaptiveSizePolicyWeight=${GC_ADAPTIVE_SIZE_POLICY_WEIGHT:-90}
   local maxMetaspaceSize=${GC_MAX_METASPACE_SIZE:-100}
+  local gcOptions="${GC_CONTAINER_OPTIONS:--XX:+UseParallelOldGC}"
 
-  echo "-XX:+UnlockExperimentalVMOptions " \
-       "-XX:+UseCGroupMemoryLimitForHeap " \
-       "-XX:+UseParallelOldGC " \
+  echo "${gcOptions}" \
+       "-XX:+UnlockExperimentalVMOptions "\
+       "-XX:+UseCGroupMemoryLimitForHeap "\
        "-XX:MinHeapFreeRatio=${minHeapFreeRatio} "\
        "-XX:MaxHeapFreeRatio=${maxHeapFreeRatio} "\
        "-XX:GCTimeRatio=${timeRatio} "\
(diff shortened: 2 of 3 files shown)


jmtd:
I should probably add documentation in os-java-run/module.yaml

os-java-run/module.yaml
```diff
@@ -38,3 +38,6 @@ envs:
     - name: GC_MAX_METASPACE_SIZE
       description: The maximum metaspace size.
       example: "100"
+    - name: GC_OPTIONS
```
rcernich: Is this descriptive enough?  It seems like it's pretty generic and could apply to any GC setting.  Also, are there cases where some GC options do or do not apply to a specific GC algorithm, e.g. does AdaptiveSizePolicyWeight apply to GCG1?

jmtd: > Is this descriptive enough?

Maybe not. As it happens, users could stuff any argument they like into that variable, but we don't advertise it for that purpose. (that's the origins of the generic name, though.) I am happy to rename GC_OPTIONS to something more specific to GC choice, but I'd like to avoid having to hardcode a set of valid GC choices in the shell scripts since that would couple us tightly to specific JVM vendor/version. The other way to go would be to re-document this as a generic argument-passing mechanism that can do GC choices, or anything else. If we went that way we could deprecate the more specific GC variables.

> Also, are there cases where some GC options do or do not apply to a specific GC algorithm, e.g. does AdaptiveSizePolicyWeight apply to GCG1?

…good question. I haven't been able to answer it in the last 20 minutes of trying. I'm going to doorstep a JVM guy when he's back in the office :)

jmtd: > I'm going to doorstep a JVM guy when he's back in the office :)

My luck is in - there's two OpenJDK devs in the office this week!

At least `adaptiveSizePolicyWeight`, `MinHeapFreeRatio` and `MaxHeapFreeRatio` are only relevant to the single-threaded and parallel GCs. They are, however, benign for the other GCs. There currently is no warning message from the JVM printed if you specify them and a GC that does not use them, but one could appear in future JVM releases.

One option would be to deprecate and remove `GC_MIN_HEAP_FREE_RATIO`, etc., folding them all into a generic `GC_OPTIONS` variable that a user can specify in its entirety, alongside the choice of GC. This is my preferred option, but does involve deprecating/removing existing configuration variables that we have exposed, so I don't know whether that'll fly.

Another option would be to encode in our shell script information about which options are relevant to which GCs, but this would tightly couple the shell script launcher to OpenJDK and the versions we are concerned with specifically, which I think would be a bad idea, a potential maintenance burden and we'd risk lagging behind OpenJDK changes.

There are probably options in the middle too.

On a related note, although the driver for this was to specify the G1GC I believe, apparently testing/benchmarking of Shenandoah for these applications is looking very promising and we could see interest from people in using it for cloud container apps.

rcernich: As `GC_` is the common prefix we're using for all of these variables, `GC_OPTIONS` seemed too vague to me, as I think all of the `GC_` variables are options.  Something like `GC_COLLECTOR_OPTIONS` seemed like a better choice to me, where that would be defaulted to the existing set of collector options.  If a user overrides that, we just use what they specified.

My second question was geared more toward whether or not we needed to change any of our other settings if the user chooses a different collector, examples being: ParallelGCThreads, MinHeapFreeRatio, MaxHeapFreeRatio, GCTimeRatio, AdaptiveSizePolicyWeight, MaxMetaspaceSize.

rcernich: Were we going to change the name of the env?

jmtd: I've gone with `GC_COLLECTOR_OPTIONS` as you suggest. I was a little uncomfortable since the C in GC is "collector" so this is "Garbage Collector Collector Options" but I couldn't come up with anything better.

> My second question was geared more toward whether or not we needed to change any of our other settings if the user chooses a different collector, examples being: ParallelGCThreads, MinHeapFreeRatio, MaxHeapFreeRatio, GCTimeRatio, AdaptiveSizePolicyWeight, MaxMetaspaceSize.

Sorry yes I understood the question but answered it so poorly that it looks like I didn't ;)

The answer is: some of those options will not have any effect if the user switches GC. But setting them will have no ill effects. How could we handle this?

1. as in the current state of this PR: we don't. 
2. we track which GC the user has selected and hard-code which options are relevant to which GC in our scripts. I really want to avoid this.
3. we deprecate (and eventually remove) those options entirely,  and use something like `GC_OPTIONS` as a flat all-gc-options-go-here variable; and the user is responsible for choosing an appropriate selection of compatible options.

My preference is 1 3 2. The main drawback of 1) is user feedback if they specify something non-sensical. However, at the moment, if they override only `GC_COLLECTOR_OPTIONS` and nothing else, *we* specify non-sensical arguments (since we define defaults for e.g. MinHeapFreeRatio and pass them regardless). I think the JDK might (and should) warn about incompatible options in the future, but right now it doesn't.

rcernich: I'm good with that.

PR #269, opened 2018-07-16
CLOUD-2706 update JSON logging to capture server boot

luck3y:
https://issues.jboss.org/browse/CLOUD-2706
Signed-off-by: Ken Wills <kwills@redhat.com>

See also https://github.com/jboss-container-images/jboss-eap-modules/pull/17 for an associated, required change to support this behavior on EAP CD.

### DIFF
--- a/os-eap-logging/added/logging.properties
+++ b/os-eap-logging/added/logging.properties
@@ -32,7 +32,7 @@ handlers=CONSOLE
 
 handler.CONSOLE=org.jboss.logmanager.handlers.ConsoleHandler
 handler.CONSOLE.level=ALL
-handler.CONSOLE.formatter=COLOR-PATTERN
+handler.CONSOLE.formatter=##CONSOLE-FORMATTER##
 handler.CONSOLE.properties=autoFlush,target,enabled
 handler.CONSOLE.autoFlush=true
 handler.CONSOLE.target=SYSTEM_OUT
--- a/os-eap7-launch/added/launch/json_logging.sh
+++ b/os-eap7-launch/added/launch/json_logging.sh
@@ -7,7 +7,9 @@ function configure_json_logging() {
 
   if [ "${ENABLE_JSON_LOGGING^^}" == "TRUE" ]; then
     sed -i 's|##CONSOLE-FORMATTER##|OPENSHIFT|' $CONFIG_FILE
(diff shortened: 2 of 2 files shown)


os-eap-logging/added/logging.properties
```diff
@@ -32,7 +32,7 @@ handlers=CONSOLE
 
 handler.CONSOLE=org.jboss.logmanager.handlers.ConsoleHandler
 handler.CONSOLE.level=ALL
-handler.CONSOLE.formatter=COLOR-PATTERN
+handler.CONSOLE.formatter=##CONSOLE-FORMATTER##
```
bstansberry: Having this disconnected from the related script is a problem, as this module is used in other images, and those images might not use os-eap7-launch. For sure EAP 6.4 doesn't, probably other products too.

bstansberry: BTW I don't mean to suggest this file must move to be with the script. I'm not suggesting a solution, just a pointing out a problem. :)

A possible solution is to limit this fix to EAP CD and EAP on OS 7.2 which can use the module from jboss-eap-modules.

luck3y: Yeah, my intention there was to do the substitution if it was there, and if it wasn't it would just be a no-op. Other images can add the ##CONSOLE-FORMATTER## as they see fit.

I'll have a quick look in cct_module and see if there are any other examples out there, then ping those people with a heads up.

luck3y: Actually never mind, sorry, I see what you mean about the file being in os-eap-logging, but the substitution being in os-eap7-launch, I'll have a look at what we can do in os-eap-logging ...

bstansberry: The problem is the logging.properties file is no longer usable as is because the handler.CONSOLE.formatter=##CONSOLE-FORMATTER## is invalid.

luck3y: @bstansberry Yeah, got it, theres a stuck pending comment from me just above your last message :(

luck3y: Sigh, the GH UI thought I'd started a review ...

luck3y: So thinking about this a little more, we should fork the cct_module os-eap-logging module to jboss-eap-modules, and install the script that does the substitution into launch/ and ensure it gets run. We can deprecate the old one, but it would still work.

PR #276, opened 2018-08-03
[CLOUD-2767] setting bigger timeout for recovery scan

ochaloup:
https://issues.jboss.org/browse/CLOUD-2767

Setting the socket timeout of the recovery request to 30 minutes while expecting whole recovery process (both phases) will be run in such time.
From the perspective of the script we need to ensure that scan is completly finished and only then server shutdown is requested.


luck3y:
@ochaloup I can't start reviews in cct_module, but +1 from me.

PR #281, opened 2018-08-14
[CLOUD-1505] - drop mping socket binding from config

luck3y:
https://issues.jboss.org/browse/CLOUD-1502
Signed-off-by: Ken Wills <kwills@redhat.com>

### DIFF
--- a/os-eap7-launch/added/launch/ha.sh
+++ b/os-eap7-launch/added/launch/ha.sh
@@ -84,7 +84,7 @@ function configure_ha() {
   fi
 
   local ping_protocol=${JGROUPS_PING_PROTOCOL:-openshift.KUBE_PING}
-  local ping_protocol_element="<protocol type=\"${ping_protocol}\" socket-binding=\"jgroups-mping\"/>"
+  local ping_protocol_element="<protocol type=\"${ping_protocol}\"/>"
   validate_ping_protocol "${ping_protocol}" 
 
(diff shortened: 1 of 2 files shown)


bstansberry:
This is a breaking change.  JGROUPS_PING_PROTOCOL can be anything, not limited to KUBE_PING or DNS_PING, and perhaps that anything needs the socket binding. MPING of course being the likely candidate.

It might be possible for MPING to work: https://blog.openshift.com/service-discovery-openshift-using-multicast and https://docs.openshift.com/container-platform/3.10/admin_guide/managing_networking.html#admin-guide-networking-multicast

luck3y:
Aha, OK, that seems like a good reason to keep it in then :)

luck3y:
So I think we have a couple of options here...

- Leave the socket binding for mping in socket-bindings, then:

Either:
 1) just leave it on in protocol ... but this seems bad if you did want to use some other protocol, even if it wasn't MC, it would still get the mping SB.
 2) disable it for 'known protocols', like KUBE_PING and DNS_PING, add it back for anything else (again, bad if you have something that isn't MC, but uses a SB)
 3) Add a JGROUPS_PING_PROTOCOL_SOCKET_BINDING that if it's set, we use, if not we fall back to #2 / #1.

Of course, doing #3 sounds like a good idea, but then we're going to need to provide a way to add a socket binding ... but we should probably have that anyway, IMO.

luck3y:
Closing this, we'll take a slightly different approach in https://github.com/jboss-container-images/jboss-eap-modules/pull/20

PR #284, opened 2018-08-29
[CLOUD-2406] Fix and improve Elytron configuration integration

bstansberry:
https://issues.jboss.org/browse/CLOUD-2406

* Default HTTPS_KEY_PASSWORD to HTTPS_PASSWORD
* Don't require HTTPS_KEYSTORE_DIR; use jboss.server.config.dir path if not set
* If HTTPS_KEYSTORE_DIR starts with '/' treat it as an absolute path, else as config model path name
* If settings are mssing, state what they are

- [X ] Pull Request title is properly formatted: `[CLOUD-XYA] Subject`
- [ X] Pull Request contains link to the JIRA issue
- [X ] Pull Request contains description of the issue
- [X ] Pull request does not include fixes for other issues than the main ticket
- [X ] Attached commits represent unit of work and are properly formatted
- [X ] You have read and agreed to the Developer Certificate of Origin (DCO) (see `CONTRIBUTING.md`)
- [X ] Every commit contains `Signed-off-by: Your Name <yourname@redhat.com>` - use `git commit -s`

### DIFF
--- a/os-eap7-launch/added/launch/elytron.sh
+++ b/os-eap7-launch/added/launch/elytron.sh
@@ -72,7 +72,17 @@ function configure_https() {
 
     https_connector="<https-listener name=\"https\" socket-binding=\"https\" ssl-context=\"LocalhostSslContext\" proxy-address-forwarding=\"true\"/>"
   elif [ -n "${HTTPS_PASSWORD}" -o -n "${HTTPS_KEYSTORE}" -o -n "${HTTPS_KEYSTORE_TYPE}" ]; then
-    echo "WARNING! Partial HTTPS configuration, the https connector WILL NOT be configured."
+    local missing_msg="WARNING! Partial HTTPS configuration, the https connector WILL NOT be configured. Missing:"
+    if [ -z "${HTTPS_PASSWORD}" ]; then
+      missing_msg="$missing_msg HTTPS_PASSWORD"
+    fi
+    if [ -z "${HTTPS_KEYSTORE}" ]; then
+      missing_msg="$missing_msg HTTPS_KEYSTORE"
+    fi
+    if [ -z "${HTTPS_KEYSTORE_TYPE}" ]; then
+      missing_msg="$missing_msg HTTPS_KEYSTORE_TYPE"
+    fi
+    echo $missing_msg
   fi
 
   sed -i "s|<!-- ##TLS## -->|${tls}|" $CONFIG_FILE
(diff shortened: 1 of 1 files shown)


os-eap7-launch/added/launch/elytron.sh
```diff
@@ -27,23 +27,42 @@ function configure_https() {
...
+      keystore_rel_to="relative-to=\"jboss.server.config.dir\""
+    elif [[ "${HTTPS_KEYSTORE_DIR}" =~ ^/ ]]; then
+      # Assume leading '/' means the value is a FS path
+      # Standard template behavior where the template sets this var to /etc/eap-secret-volume
+      keystore_path="path=\"${HTTPS_KEYSTORE_DIR}/${HTTPS_KEYSTORE}\""
+      keystore_rel_to=""
+    else
+      # Compatibility edge case. Treat no leading '/' as meaning HTTPS_KEYSTORE_DIR is the name of a config model path
+      keystore_path="path=\"${HTTPS_KEYSTORE}\""
+      keystore_rel_to="relative-to=\"${HTTPS_KEYSTORE_DIR}\""
```
tremes: I think it makes sense, however it doesn't correspond to the steps described in https://issues.jboss.org/browse/CLOUD-2406?focusedCommentId=13624783&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13624783 if I am not mistaken.

bstansberry: @tremes Right; it's different and I should comment on the JIRA why. Thanks for reminding me!

When I wrote that JIRA comment I didn't realize that our templates were setting HTTPS_KEYSTORE_DIR to /etc/eap-secret-volume. That's not a config model logical path name and that usage shows me that the primary use case for this setting was as a FS path in https.sh.  So to be user friendly having it work as a FS path here as should be supported, with the "config model logical path name" being more of an edge case for the chance that someone got this working despite CLOUD-2406.

So, we test HTTPS_KEYSTORE_DIR for leading '/' and act accordingly.

In that comment I also talk about testing HTTPS_KEYSTORE for a leading '/' and treating it as an absolute path and not doing relative-to if it has one. But, Ken and I realized that in no existing code did we have that functionality. Adding it would basically be a minor RFE. And with the handling of HTTPS_KEYSTORE_DIR already being complex, adding yet another config permutation would have just been too confusing and bug prone. Better to keep it simple and continue to treat HTTP_KEYSTORE as a filename relative to some other path.

bstansberry: I commented on the JIRA as well.

PR #296, opened 2018-09-20
[CLOUD-2542] storing txn recovery data in the jdbc object store

ochaloup:
https://issues.jboss.org/browse/CLOUD-2542

The purpose of this change is storing OpenShift txn recovery data, which was introduced by changes in CLOUD-2261, to database. The database is used when also transaction manager uses jdbc object store as part of the JBoss EAP configuration.

@rcernich would you be so kind and review this PR and could be pass this later to QA?

/cc @tomjenkinson


bstansberry:
@ochaloup is the keystore.jks file in this PR meant to be there?

The part integrating and invoking the new module looks fine.

ochaloup:
@bstansberry correct. The `keystore.jks` was only for testing purposes on my local computer and unfortunatelly I added it to the PR. I hope it's ok now. Thank you for spotting it and in general for the review!

ochaloup:
@tomjenkinson I think it's not necessary. I could create `keystore.jks` at any location on my drive that the purpose was to to import it to OpenShift. That was only my inattention to create it at the `cct_module` repo and on top of it adding it to this PR.

rcernich:
@luck3y, @bstansberry, I'm going to defer to you for dev-approved on this.  If you're okay with the changes, please add the label to the jira.  TIA

ochaloup:
@luck3y I found I need to make a change for this works in EAP6. As the tool uses hibernate the version 4.2 in EAP6 defines different API than the 5.1 in EAP7.

I need to use different jar file for the EAP7 and EAP6.
I've made a change here in the PR
https://github.com/jboss-openshift/cct_module/pull/296/commits/e0d0fcc5850f2f327d4e250caf71e49e18ad5c6e#diff-1bc477c3bbe222b9df2729b7d813a61aR16
plus then I would like to add two more PRs for `jboss-eap-modules` and template for eap 64.

* https://github.com/ochaloup/jboss-eap-modules/commit/c017628d09f274e87b2a88512f04da8d43ae1b8b
* https://github.com/ochaloup/application-templates/commit/bf595dd6d384f0ddf7b7284e7d791bb5fb048bb6

Will be such structure ok?

ochaloup:
@luck3y, @bstansberry : I've finished with productization for the txn recovery tool to be running with EAP6. Would you be so kind and review this PR?

os-partition-txnrecovery/configure.sh
```diff
@@ -1,11 +0,0 @@
-#!/bin/sh
```
tomjenkinson: Just to mention I can't see where the call to this used to be?

ochaloup: the script to be applied for particular module is defined  in descriptor `module.yaml` like here https://github.com/jboss-openshift/cct_module/blob/35018c1213fd064124b0464ac8703fd506f14548/os-partition-txnrecovery/module.yaml

ochaloup:
the PR for jboss-eap-modules related to this was crated: https://github.com/jboss-container-images/jboss-eap-modules/pull/38

PR #304, opened 2018-09-27
[CLOUD-2912] Do not set AUTO_DEPLOY_EXPLODED for SSO integration

ruromero:
Signed-off-by: Ruben Romero Montes <rromerom@redhat.com>

Fix https://issues.jboss.org/browse/CLOUD-2912

### DIFF
--- a/os-eap-sso/added/keycloak.sh
+++ b/os-eap-sso/added/keycloak.sh
@@ -135,12 +135,10 @@ function set_curl() {
 
 function enable_keycloak_deployments() {
   if [ -n "$SSO_OPENIDCONNECT_DEPLOYMENTS" ]; then
-    AUTO_DEPLOY_EXPLODED="true"
     explode_keycloak_deployments $SSO_OPENIDCONNECT_DEPLOYMENTS $OPENIDCONNECT
   fi
 
   if [ -n "$SSO_SAML_DEPLOYMENTS" ]; then
-    AUTO_DEPLOY_EXPLODED="true"
     explode_keycloak_deployments $SSO_SAML_DEPLOYMENTS $SAML
   fi
 }
@@ -155,6 +153,9 @@ function explode_keycloak_deployments() {
       unzip -o ${JBOSS_HOME}/standalone/deployments/${sso_deployment} -d ${JBOSS_HOME}/standalone/deployments/tmp
       rm -f ${JBOSS_HOME}/standalone/deployments/${sso_deployment}
       mv ${JBOSS_HOME}/standalone/deployments/tmp ${JBOSS_HOME}/standalone/deployments/${sso_deployment}
+      if [ ! -f ${JBOSS_HOME}/standalone/deployments/${sso_deployment}.dodeploy ]; then
+        touch ${JBOSS_HOME}/standalone/deployments/${sso_deployment}.dodeploy
+      fi
     fi
 
     if [ -f "${JBOSS_HOME}/standalone/deployments/${sso_deployment}/WEB-INF/web.xml" ]; then
@@ -223,9 +224,7 @@ function configure_subsystem() {
       web_xml=`read_web_dot_xml $f WEB-INF/web.xml`
       if [ -n "$web_xml" ]; then
         requested_auth_method=`echo $web_xml | xmllint --nowarning --xpath "string(//*[local-name()='auth-method'])" - | sed ':a;N;$!ba;s/\n//g' | tr -d '[:space:]'`
-
-        if [[ $requested_auth_method == "${auth_method}" ]]
-        then
+        if [[ $requested_auth_method == "${auth_method}" ]]; then
 
           if [ -z "$subsystem" ]; then
             subsystem="${keycloak_subsystem}"
@@ -436,10 +435,8 @@ function read_web_dot_xml {
   local result=
 
   if [ -d "$jarfile" ]; then
-    if [[ -n "$AUTO_DEPLOY_EXPLODED" && "$AUTO_DEPLOY_EXPLODED" == "true" ]] || [[ -n "$JAVA_OPTS_APPEND" && $JAVA_OPTS_APPEND == *"Xdebug"* ]]; then
-      if [ -e "${jarfile}/${filename}" ]; then
+    if [ -e "${jarfile}/${filename}" ]; then
         result=`cat ${jarfile}/${filename}`
-      fi
     fi
   else
     file_exists=`unzip -l "$jarfile" "$filename"`


rcernich:
@luck3y, can you take a look at this please?  My understanding is that we will modify the web.xml to properly configure sso for authentication, so we expand the war and it should be deployed as an exploded war.  If this is good for you, it's good for me.  (Some history, I believe this was added because even if the deployment is configured for sso in standalone.xml, it wasn't working correctly if the app was configured to use BASIC, something about EAP trying to use both.  Funny thing is, iirc, that this was only an issue with BASIC, and maybe just 6.4.)

luck3y:
@ruromero I'm not 100% sure I understand what the issue is here. The explode_sso_deployments has:
`
      mkdir ${JBOSS_HOME}/standalone/deployments/tmp
      unzip -o ${JBOSS_HOME}/standalone/deployments/${sso_deployment} -d ${JBOSS_HOME}/standalone/deployments/tmp
      rm -f ${JBOSS_HOME}/standalone/deployments/${sso_deployment}
      mv ${JBOSS_HOME}/standalone/deployments/tmp ${JBOSS_HOME}/standalone/deployments/${sso_deployment}
`

Would it make more sense to do this outside of deployments?

bstansberry:
This looks like a breaking behavioral change at least in some scenario, so for sure we need to be very careful. Hence this long post.

I've written to Jeff Maury about how the dev tooling uses the scanner. My expectation is they do not count on auto-deploy-exploded, but we should confirm. If they don't that eliminates a whole category of concern.

In general, auto-deploy-exploded=true is poor practice for anything but personal dev experimentation, so I'd certainly be happier if it was not being turned on except directly by the user.

AIUI the situation now is AUTO_DEPLOY_EXPLODED would only be getting set to true if one of these is true:

1. if [ -f $SECURE_DEPLOYMENTS ] || [ -f $SECURE_SAML_DEPLOYMENTS ]
2. elif [ -n "$SSO_URL" ]

and in addition one of these must be true

1. if [ -n "$SSO_OPENIDCONNECT_DEPLOYMENTS" ]
2. if [ -n "$SSO_SAML_DEPLOYMENTS" ]

So, basically only if some keycloak-related deployments are configured.  So this is not a generally breaking change.

The downside is in this case, the auto-deploy behavior will change for both the keycloak-related deployments, and any others deployed to the same server.  For the 'any others' case, IMO a breaking change like that is fine; it's an edge case and users can just set AUTO_DEPLOY_EXPLODED=true themselves if they want it, and not count on obscure keycloak integration behavior to do it.  For the keycloak-related deployments case I think it's ok too, it's just not such an edge case.


General question aside, unless I've missing something I don't think this fix is correct, as it removes turning on AUTO_DEPLOY_EXPLODED but doesn't replace that with a marker file for the deployments that get exploded. The result is those deployments won't be deployed.

ruromero:
> @ruromero Is this a new issue, or one thats been happening for some time?

This seems to be an old issue but the SSO integration in RHPAM is rather recent. That´s why it has not been reported until now

rcernich:
all, keep in mind this script is shared by eap 6.4 too, so any testing will need to be done on that image as well.

bstansberry:
@ruromero If an exploded deployment is deploying at boot without either auto-deploy-exploded=true or a marker file, there's a bug. I just tested putting an exploded deployment in the deployments dir of a CD14 zip build and it did not deploy at boot. (I'm quite confident EAP 7.1 and 6.4 are the same.)

Looking at this keycloak.sh it's a typical script that gets run by our server launch script before it starts the JVM. So these deployments listed in SSO_OPENIDCONNECT_DEPLOYMENTS or SSO_SAML_DEPLOYMENTS will be exploded at boot time. Without the marker they will not deploy.

ruromero:
@bstansberry you're totally right! I missed that in the RHPAM images there was already a .dodeploy file.
I have updated the script accordingly. Thanks for your comments.

os-eap-sso/added/keycloak.sh
```diff
@@ -436,10 +435,8 @@ function read_web_dot_xml {
   local result=
 
   if [ -d "$jarfile" ]; then
-    if [[ -n "$AUTO_DEPLOY_EXPLODED" && "$AUTO_DEPLOY_EXPLODED" == "true" ]] || [[ -n "$JAVA_OPTS_APPEND" && $JAVA_OPTS_APPEND == *"Xdebug"* ]]; then
-      if [ -e "${jarfile}/${filename}" ]; then
+    if [ -e "${jarfile}/${filename}" ]; then
```
bstansberry: @iankko I think this change is ok but I wanted to point it out to you in case you thought otherwise.

AIUI what this function does is it provides deployment descriptor content to the configure_subsystem function, which then checks the auth-method, and if it matches KEYCLOAK or SAML, then that factors the deployment into the configuration of the subsystem.

My impression is the point of the removed L439 was to avoid looking into exploded deployments unless some env vars were set in such a way that it was thought that the deployment would get deployed.  IOW ignore deployments that won't be deployed. So removing L439 means there's a chance some deployment gets processed that wouldn't be deployed. Which seems like an edge case.

Even without the CLOUD-2912 problem, the old L439 wasn't quite right as it was not accounting for the case where $jarfile would be deployed because a ${jarfile}.dodeploy marker file exists.  If this overall PR went in but you wanted the L439 check preserved, then that would need to be corrected, since "$AUTO_DEPLOY_EXPLODED" is no longer likely to be "true".

But if ignoring some deployments isn't important, then the change here is simpler.

bstansberry:
My above question for Jan aside, this LGTM. Sorry for forgetting to come back to this.

ruromero:
> My above question for Jan aside, this LGTM. Sorry for forgetting to come back to this.

The PR still needs to be approved, is there anything left for me to do?
CC: @luck3y @rcernich @bstansberry 
Thanks

rcernich:
@ruromero, I leave that question to @bstansberry and @luck3y.  That said, before it can be merged the issue needs QE approval, so you'll need to make sure you provide them with some test images.

os-eap-sso/added/keycloak.sh
iankko: @bstansberry 

Hey Brian,

`My impression is the point of the removed L439 was to avoid looking into exploded deployments unless some env vars were set in such a way that it was thought that the deployment would get deployed. IOW ignore deployments that won't be deployed. So removing L439 means there's a chance some deployment gets processed that wouldn't be deployed. Which seems like an edge case.`

FWICT looking at the code in question, yes, that's probably the explanation. Other PoV / reason why it was writtent, as it wast, might be, we don't want to process only deployments, that will be truly deployment, but rather pragmatical approach - if the file is already expanded in plaintext (exploded), read `web.xml` from it. If it isn't yet, unzip the the archive, and preprocess it. IOW speed-up the `web.xml` retrieval.

`Even without the CLOUD-2912 problem, the old L439 wasn't quite right as it was not accounting for the case where $jarfile would be deployed because a ${jarfile}.dodeploy marker file exists. If this overall PR went in but you wanted the L439 check preserved, then that would need to be corrected, since "$AUTO_DEPLOY_EXPLODED" is no longer likely to be "true".`

Agree, `AUTO_DEPLOY_EXPLODED` shouldn't default to `true` afaict, per:

- [1] https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.1/html-single/configuration_guide/#reference_deploy_scanner_attr or
- [2] https://docs.jboss.org/author/display/AS71/Deployment+Scanner+configuration

From [2] `Setting this to 'true' is not recommended for anything but basic development scenarios, as there is no way to ensure that deployment will not occur in the middle of changes to the content. ` => defaulting to `false` seems reasonable to me.

The added [line in PR](https://github.com/jboss-openshift/cct_module/pull/304/files#diff-fce22065a6ac781b114c39c8eb087de0R157) seems like functional alternative of setting `AUTO_DEPLOY_EXPLODED=true`, without the possibly negative effect of causing re-deployments.

The proposed PR LGTM (from code review PoV).

@ruromero @bstansberry @rcernich @luck3y Given the testing images won't show some regression (see comment from Rob about providing them below), ACK for the change.

bstansberry:
@ruromero I marked CLOUD-2912 as dev-approved, so it just needs qe-approved to merge.

PR #309, opened 2018-10-30
[CLOUD-2903] Unable to define datasource properties

rimolive:
Add 'is-same-rm-override' configuration when Oracle XA Datasource is set. JIRA Issue: **https://issues.jboss.org/browse/CLOUD-2903**

- [X] Pull Request title is properly formatted: `[CLOUD-XYA] Subject`
- [X] Pull Request contains link to the JIRA issue
- [X] Pull Request contains description of the issue
- [X] Pull request does not include fixes for other issues than the main ticket
- [X] Attached commits represent unit of work and are properly formatted
- [X] You have read and agreed to the Developer Certificate of Origin (DCO) (see `CONTRIBUTING.md`)
- [X] Every commit contains `Signed-off-by: Your Name <yourname@redhat.com>` - use `git commit -s`


os-eap-launch/added/launch/datasource-common.sh
```diff
@@ -261,6 +261,13 @@ function generate_external_datasource() {
       ds="$ds
              <max-pool-size>$max_pool_size</max-pool-size>"
     fi
+
+    # CLOUD-2903: For Oracle XA Datasources, this configuration is required
+    if [ "$driver" == "oracle" ]; then
```
rcernich: We shouldn't have DB specific code in these modules. This should be accommodated in a generic fashion, so it can be configured for any type of DB

rimolive: I totally agree with that, although I only know about this configuration because of Oracle datasources, not sure how useful would be for other datasources.

Anyway, would a PREFIX_IS_SAME_RM_OVERRIDE env var fits better in this situation?

rcernich: If you mean `<prefix>_IS_SAME_RM_OVERRIDE`, then yes, that sounds reasonable.  @luck3y, @bstansberry, opinions?

bstansberry: Sounds fine.

I'm not a fan of endless env vars for every possible use case, but this datasource use case is already so heavily configurable via env vars that adding one more doesn't matter much.  For sure if we're adding another it needs to work like the rest, i.e. <prefix>_IS_SAME_OVERRIDE.

bstansberry: Definitely checking for ""oracle" is no good since AFAIK that string has no meaning to us. It's not one of our standard driver name strings that turns on specific behavior.

errantepiphany:
@rimolive , @bstansberry , @rcernich : What is the status of this?

errantepiphany:
@luck3y  ^^

errantepiphany:
Discussion underway on the jira: https://issues.jboss.org/browse/CLOUD-2903

PR #311, opened 2018-11-05
[CLOUD-3009] correcting path to scl-enable-maven in maven s2i script

ochaloup:
https://issues.jboss.org/browse/CLOUD-3009

/cc @luck3y


rcernich:
The implementation is correct.  Please see https://github.com/jboss-openshift/cct_module/blob/master/jboss/container/maven/default/bash/configure.sh#L19

PR #312, opened 2018-11-05
CLOUD-2689: jboss.container.java.certs module

jmtd:
This module enables the trusting of CA certificates provided to
running containers by OpenShift.

At build-time: various system PKI paths are made world writeable-by-all,
in order that the random-UID runtime user can insert CA certificates
into the trusted store

At run-time: copy the CA file into place and invoke /usr/bin/update-ca-trust
To generate derivative trust stores (e.g. OpenSSL & Java JKS)

https://issues.jboss.org/browse/CLOUD-2689

### DIFF
--- a/jboss/container/java/certs/bash/artifacts/trust-ose-cert.sh
+++ b/jboss/container/java/certs/bash/artifacts/trust-ose-cert.sh
@@ -2,8 +2,8 @@
 set -u
 set -e
 
-cp /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt \
-	/etc/pki/ca-trust/source/anchors/
-/usr/bin/update-ca-trust
-
-
+ca=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
+if test -r "$ca" ; then
+    cp "$ca" /etc/pki/ca-trust/source/anchors/
+    /usr/bin/update-ca-trust
+fi
--- a/jboss/container/java/certs/bash/artifacts/trust-ose-cert.sh
+++ b/jboss/container/java/certs/bash/artifacts/trust-ose-cert.sh
@@ -2,9 +2,7 @@
 set -e
 
 if [ -z "$JAVA_DISABLE_TRUST_OPENSHIFT_CA" ]; then
-    ca=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
-    if test -r "$ca" ; then
-        cp "$ca" /etc/pki/ca-trust/source/anchors/
+    if test -r "/etc/pki/ca-trust/source/anchors/ose-service-ca.crt" ; then
         /usr/bin/update-ca-trust
     fi
 fi
--- a/jboss/container/java/certs/bash/artifacts/trust-ose-cert.sh
+++ b/jboss/container/java/certs/bash/artifacts/trust-ose-cert.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+set -u
+set -e
+
+cp /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt \
+	/etc/pki/ca-trust/source/anchors/
+/usr/bin/update-ca-trust
+
+
--- a/jboss/container/java/certs/bash/artifacts/trust-ose-cert.sh
+++ b/jboss/container/java/certs/bash/artifacts/trust-ose-cert.sh
@@ -1,9 +1,10 @@
 #!/bin/sh
-set -u
 set -e
 
-ca=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
-if test -r "$ca" ; then
-    cp "$ca" /etc/pki/ca-trust/source/anchors/
-    /usr/bin/update-ca-trust
+if [ -z "$JAVA_DISABLE_TRUST_OPENSHIFT_CA" ]; then
+    ca=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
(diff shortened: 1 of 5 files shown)


jboss/container/java/certs/bash/configure.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+# Configure module
+set -e
+
+SCRIPT_DIR=$(dirname $0)
+ARTIFACTS_DIR=${SCRIPT_DIR}/artifacts
+
+# necessary for random UID user to run update-ca-certs
```
luck3y: Is it possible to chmod these back afterwards? it seems odd to leave them 777.

jmtd: No; because the chmod is happening in the build stage but the write access (via update-ca-trust) is happening at runtime.

It's

jboss/container/java/certs/bash/artifacts/trust-ose-cert.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+set -e
+
+if [ -z "$JAVA_DISABLE_TRUST_OPENSHIFT_CA" ]; then
```
maschmid: note that there are 2 different CA...  the ca.crt , which I would call "the openshfit CA",  and the service-ca.crt is a very specific one  (https://docs.openshift.com/online/dev_guide/secrets.html#service-serving-certificate-secrets ) 

I'd prefer if we (optionally) added both.. However, even if we don't, we should make it clear in the names of the env and comments which CA is being (or not being) added.

jmtd: agreed

jboss/container/java/certs/bash/configure.sh
jmtd: checking

jmtd: Yes g+w is sufficient. The threat model doesn't really change (same user in both cases) but it feels nicer to not have it as wide as 777. I'll commit this.

jboss/container/java/certs/bash/artifacts/trust-ose-cert.sh
```diff
@@ -0,0 +1,10 @@
...
+    ca=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
+    if test -r "$ca" ; then
+        cp "$ca" /etc/pki/ca-trust/source/anchors/
```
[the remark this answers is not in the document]
jmtd: This seems to work fine. Thanks for the suggestion!

jboss/container/java/certs/bash/configure.sh
rcernich: Do we need to update the permissions if we're creating the link during image build?  Also, can the pod user run the update script without special privs?  Will the update script fail if the link is broken (i.e. the target file doesn't exist)?  Do we need to chmod the link, to make sure the perms are correct there (I had this problem in other places, where the link was root:root, which required an explicit `chown -h`.

rcernich: If all of the above works out okay (i.e. link has correct owner, runtime user can invoke the update script, not having to update perms on the directories), then I think that eliminates and problems.

jmtd: > Do we need to update the permissions if we're creating the link during image build?

Yes, because we are widening the permissions on the locations to which the derivative TLS bundles are written to. The "update-ca-certs" scripts reads all the certs in a certain path (including the OpenShift one via symlink), it only needs read access to those. But it generates the OpenSSL and JKS stores, and needs write permission to put them in the system location.

> Also, can the pod user run the update script without special privs?

Yes if the permission change has taken place, that's the only additional priv required.

> Will the update script fail if the link is broken (i.e. the target file doesn't exist)? 

No, silently ignored.

>  Do we need to chmod the link, to make sure the perms are correct there (I had this problem in other places, where the link was root:root, which required an explicit chown -h.

I don't think so, but in my testing I've just been launching the containers with `docker` and volume-mounting a certificate with `-v`. It's not impossible that OpenShfit does something subtly different which could expose a problem here.

rcernich: make sure you try running the container with something like `-u 10010000`

rcernich: whoops, 10010000:0

jmtd: @rcernich the test harness I wrote to check this stuff out, I published [here](https://github.com/jmtd/redhat-openjdk-18-openshift-image/tree/cloud-2689-certificates-TEST). In particular see [test.sh line 1])(https://github.com/jmtd/redhat-openjdk-18-openshift-image/blob/cloud-2689-certificates-TEST/test.sh#L1): each invocation of the test harness uses a random uid, just like openshift.

rcernich: Yes, but OpenShift also runs with group 0.

jmtd: Ah I missed the group spec! I'll adjust my harness and re-test.

jmtd: [fixed](https://github.com/jmtd/redhat-openjdk-18-openshift-image/commit/f04e2136c9088eafadbb767fae1226f428b691fa) and re-tested locally; seems fine

jmtd:
I don't think we need this anymore: I think OpenShift (4.1? 4.2?) is handling this via an operator.

PR #314, opened 2018-11-13
CLOUD-3013 add basic prometheus modules

rcernich:
Signed-off-by: rcernich <rcernich@redhat.com>

### DIFF
--- a/jboss/container/eap/prometheus/config/6.4/artifacts/opt/jboss/container/prometheus/etc/jmx-exporter-config.yaml
+++ b/jboss/container/eap/prometheus/config/6.4/artifacts/opt/jboss/container/prometheus/etc/jmx-exporter-config.yaml
@@ -1,0 +1,343 @@
+#
+# Copyright 2018 Red Hat, Inc. and/or its affiliates
+# and other contributors as indicated by the @author tags.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+---
+lowercaseOutputName: true
+lowercaseOutputLabelNames: true
+whitelistObjectNames:
+- "jboss.as:deployment=*,subsystem=ejb3,*"
+- "jboss.as:deployment=*,subsystem=web"
+- "jboss.as:deployment=*,subsystem=web,servlet=*"
+- "jboss.as:deployment=*,subdeployment=*,subsystem=ejb3,*"
+- "jboss.as:deployment=*,subdeployment=*,subsystem=web"
+- "jboss.as:deployment=*,subdeployment=*,subsystem=web,servlet=*"
+- "jboss.as:subsystem=datasources,data-source=*,statistics=jdbc"
+- "jboss.as:subsystem=datasources,data-source=*,statistics=pool"
+- "jboss.as:subsystem=datasources,xa-data-source=*,statistics=jdbc"
+- "jboss.as:subsystem=datasources,xa-data-source=*,statistics=pool"
+- "jboss.as:subsystem=messaging,hornetq-server=*,jms-queue=*"
+- "jboss.as:subsystem=messaging,hornetq-server=*,jms-topic=*"
+- "jboss.as:subsystem=transactions"
+blacklistObjectNames:
+# handled by agent's default exporter
+- "java.lang:*"
+rules:
+# EAP Specific
+# Web Metrics
+# Sessions
+- pattern: "jboss.as<deployment=(.+), *subsystem=web><>(active_sessions|max_active_sessions): (.+)"
+  name: web_$2
+  help: Web Session Metrics $2
+  value: $3
+  valueFactor: 1
+  labels:
+    deployment: $1
+    subdeployment: $1
+  attrNameSnakeCase: true
+  type: GAUGE
+- pattern: "jboss.as<deployment=(.+), *subdeployment=(.+), *subsystem=web><>(active_sessions|max_active_sessions): (.+)"
+  name: web_$3
+  help: Web Session Metrics $3
+  value: $4
+  valueFactor: 1
+  labels:
+    deployment: $1
+    subdeployment: $2
+  attrNameSnakeCase: true
+  type: GAUGE
+- pattern: "jboss.as<deployment=(.+), *subsystem=web><>(session_avg_alive_time|session_max_alive_time): (.+)"
+  name: web_$2_seconds
+  help: Web Session Metrics $2
+  value: $3
+  valueFactor: .001
+  labels:
+    deployment: $1
+    subdeployment: $1
+  attrNameSnakeCase: true
+  type: GAUGE
+- pattern: "jboss.as<deployment=(.+), *subdeployment=(.+), *subsystem=web><>(session_avg_alive_time|session_max_alive_time): (.+)"
+  name: web_$3_seconds
+  help: Web Session Metrics $3
+  value: $4
+  valueFactor: .001
+  labels:
+    deployment: $1
+    subdeployment: $2
+  attrNameSnakeCase: true
+  type: GAUGE
+- pattern: "jboss.as<deployment=(.+), *subsystem=web><>(duplicated_session_ids|expired_sessions|rejected_sessions|sessions_created): (.+)"
+  name: web_$2_total
+  help: Web Session Metrics $2
+  value: $3
+  valueFactor: 1
+  labels:
+    deployment: $1
+    subdeployment: $1
+  attrNameSnakeCase: true
+  type: COUNTER
+- pattern: "jboss.as<deployment=(.+), *subdeployment=(.+), *subsystem=web><>(duplicated_session_ids|expired_sessions|rejected_sessions|sessions_created): (.+)"
+  name: web_$3_total
+  help: Web Session Metrics $3
+  value: $4
+  valueFactor: 1
+  labels:
+    deployment: $1
+    subdeployment: $2
+  attrNameSnakeCase: true
+  type: COUNTER
+# Servlets
+- pattern: "jboss.as<deployment=(.+), *subsystem=web, *servlet=(.+)><>(load_time|max_time|min_time): (.+)"
+  name: web_servlet_$3_seconds
+  help: Web Servlet Metrics $3
+  value: $4
+  valueFactor: .001
+  labels:
+    deployment: $1
+    subdeployment: $1
+    servlet: $2
+  attrNameSnakeCase: true
+  type: GAUGE
+- pattern: "jboss.as<deployment=(.+), *subdeployment=(.+), *subsystem=web, *servlet=(.+)><>(load_time|max_time|min_time): (.+)"
+  name: web_servlet_$4_seconds
+  help: Web Servlet Metrics $4
+  value: $5
+  valueFactor: .001
+  labels:
+    deployment: $1
+    subdeployment: $2
+    servlet: $3
+  attrNameSnakeCase: true
+  type: GAUGE
+- pattern: "jboss.as<deployment=(.+), *subsystem=web, *servlet=(.+)><>(processing_time): (.+)"
+  name: web_servlet_$3_seconds_total
+  help: Web Servlet Metrics $3
+  value: $4
+  valueFactor: .001
+  labels:
+    deployment: $1
+    subdeployment: $1
+    servlet: $2
+  attrNameSnakeCase: true
+  type: COUNTER
+- pattern: "jboss.as<deployment=(.+), *subdeployment=(.+), *subsystem=web, *servlet=(.+)><>(processing_time): (.+)"
+  name: web_servlet_$4_seconds_total
+  help: Web Servlet Metrics $4
+  value: $5
+  valueFactor: .001
+  labels:
+    deployment: $1
+    subdeployment: $2
+    servlet: $3
+  attrNameSnakeCase: true
+  type: COUNTER
+- pattern: "jboss.as<deployment=(.+), *subsystem=web, *servlet=(.+)><>(request_count): (.+)"
+  name: web_servlet_$3_total
+  help: Web Servlet Metrics $3
+  value: $4
+  valueFactor: 1
+  labels:
+    deployment: $1
+    subdeployment: $1
+    servlet: $2
+  attrNameSnakeCase: true
+  type: COUNTER
+- pattern: "jboss.as<deployment=(.+), *subdeployment=(.+), *subsystem=web, *servlet=(.+)><>(request_count): (.+)"
+  name: web_servlet_$4_total
+  help: Web Servlet Metrics $4
+  value: $5
+  valueFactor: 1
+  labels:
+    deployment: $1
+    subdeployment: $2
+    servlet: $3
+  attrNameSnakeCase: true
+  type: COUNTER
+
+# EJB
+- pattern: "jboss.as<deployment=(.+), *subsystem=ejb3, *(.+)=(.+)><>(peak_concurrent_invocations|pool_available_count|pool_current_size|pool_max_size): (.+)"
+  name: ejb3_$4
+  help: EJB Metrics $4
+  value: $5
+  valueFactor: 1
+  labels:
+    deployment: $1
+    subdeployment: $1
+    type: $2
+    name: $3
+  attrNameSnakeCase: true
+  type: GAUGE
+- pattern: "jboss.as<deployment=(.+), *subdeployment=(.+), *subsystem=ejb3, *(.+)=(.+)><>(peak_concurrent_invocations|pool_available_count|pool_current_size|pool_max_size): (.+)"
+  name: ejb3_$5
+  help: EJB Metrics $5
+  value: $6
+  valueFactor: 1
+  labels:
+    deployment: $1
+    subdeployment: $2
+    type: $3
+    name: $4
+  attrNameSnakeCase: true
+  type: GAUGE
+- pattern: "jboss.as<deployment=(.+), *subsystem=ejb3, *(.+)=(.+)><>(invocations|pool_create_count|pool_remove_count): (.+)"
+  name: ejb3_$4_total
+  help: EJB Metrics $4
+  value: $5
+  valueFactor: 1
+  labels:
+    deployment: $1
+    subdeployment: $1
+    type: $2
+    name: $3
+  attrNameSnakeCase: true
+  type: COUNTER
+- pattern: "jboss.as<deployment=(.+), *subdeployment=(.+), *subsystem=ejb3, *(.+)=(.+)><>(invocations|pool_create_count|pool_remove_count): (.+)"
+  name: ejb3_$5_total
+  help: EJB Metrics $5
+  value: $6
+  valueFactor: 1
+  labels:
+    deployment: $1
+    subdeployment: $2
+    type: $3
+    name: $4
+  attrNameSnakeCase: true
+  type: COUNTER
+- pattern: "jboss.as<deployment=(.+), *subsystem=ejb3, *(.+)=(.+)><>(execution_time|wait_time): (.+)"
+  name: ejb3_$4_seconds_total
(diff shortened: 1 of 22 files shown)


jboss/container/eap/prometheus/config/7.1/artifacts/opt/jboss/container/prometheus/etc/jmx-exporter-config.yaml
```diff
@@ -0,0 +1,354 @@
...
+# EJB
+- pattern: "jboss.as<deployment=(.+), *subsystem=ejb3, *(.+)=(.+)><>(peak_concurrent_invocations|pool_available_count|pool_current_size|pool_max_size): (.+)"
+  name: ejbs_$4
+  help: EJB Metrics $4
+  value: $5
+  valueFactor: 1
+  labels:
+    deployment: $1
+    type: $2
+    bean: $3
```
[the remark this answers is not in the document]
jmesnil: I agree that it would be better to be consistent with other patterns and have different labels for the 4 children of http://wildscribe.github.io/WildFly/14.0/deployment/subsystem/ejb3/index.html

I am currently adding to the metrics exposed by the WildFly subsystem in https://github.com/wildfly/wildfly/pull/11847 so that it maps to the labels you propose. Once this PR is merged, the metric [message-driven-bean's peak-concurrent-invocations](
http://wildscribe.github.io/WildFly/14.0/deployment/subsystem/ejb3/message-driven-bean/index.html#attr-peak-concurrent-invocations) will be tagged with:

```
deployment=example.war
subsystem=ejb3
message-driven-bean= X
attribute=peak-concurrent-invocations
```

The name of the Prometheus metric is a mouthful though: `deployment_example_war_subsystem_ejb3_message_driven_bean_x_peak_concurrent_invocations`

jboss/container/eap/prometheus/config/6.4/artifacts/opt/jboss/container/prometheus/etc/jmx-exporter-config.yaml
```diff
@@ -0,0 +1,354 @@
...
+  value: $4
+  valueFactor: 1
+  labels:
+    deployment: $1
+    subdeployment: $2
+  attrNameSnakeCase: true
+  type: COUNTER
+# Servlets
+- pattern: "jboss.as<deployment=(.+), *subsystem=web, *servlet=(.+)><>(load_time|max_time|min_time): (.+)"
+  name: web_servlets_$3_seconds
```
jmesnil: Should the name be unique?

Affect, the name is composed of the attribute name only (`load_time`, 'max_time` or `min_time`).
How would you distinguish the metric `web_servlets_max_time_seconds` between different servlet (in different deployment)?

The Microprofile Metrics mandates that the metric name is unique in a given scope. So I ended up with a metric name composed of the resource address appended with the attribute name.

jboss/container/eap/prometheus/config/7.1/artifacts/opt/jboss/container/prometheus/etc/jmx-exporter-config.yaml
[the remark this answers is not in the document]
rcernich: I can make these changes, and `$2: $3` is a legal substitution.

That said, @jmesnil, I think it's not very prometheusy to include labely things in the metric name.  I think the idea here is that the metric name would be `ejb3_peak_concurrent_invocations` with the distinction for deployment, bean type, and bean name being specified on the labels.  I think the idea here is that this allows users to more easily aggregate metrics, in this case `peak_concurrent_invocations` for ejbs (`ejb3_`), assuming I correctly understand the intent.

jboss/container/eap/prometheus/config/6.4/artifacts/opt/jboss/container/prometheus/etc/jmx-exporter-config.yaml
bstansberry: It sounds like we have a conflict between the spec requirement and the "prometheusy" thing @rcernich mentioned above. I get the point of the "prometheusy" approach; it's more flexible.

jmesnil: I agree that the MP spec is too restrictive (uniqueness should be checked with name + tags).

I'll raise the issue on https://github.com/eclipse/microprofile-metrics

rcernich:
I've updated the PR to address the comments and better align with the future microprofile-metrics implementation.  The names are now of the form: `subsystem[_path]_attribute[_units]`, e.g. `undertow_active_sessions` or `undertow_servlet_max_request_time_seconds`.  I've also added a `subsystem` label, as well as rearranged subdeployment to parent and deployment instead of deployment and subdeployment.  For specific resource types, the labels are now `<type>: <name>`, e.g. `data-source: ExampleDS`.

bstansberry:
@rcernich Thanks!  You foolishly gave consideration to our thinking, which inspired more thinking. Here's a chat Jeff and I just had in the WF slack channel. WDYT?

tl;dr; from @jmesnil 
"I would use <subsystem>_<type*>_<attribute_name> with as much type as require to disambiguate
so, undertow_unique_metric
undertow_listener_request_count
undertow_servlet_request_count"

Brian Stansberry [10:34 AM]
@jmesnil so re metric names, aiui the advantage of fairly generic names is they can be aggregated using labels

undertow_request_count can be aggregrated across the many metrics with different labels (the different deployment / servlet resources)
but i don’t know if there’s any good “best general fit”

Jeff Mesnil [10:37 AM]
I agree with you, prometheus approach is fine

Brian Stansberry [10:37 AM]
perhaps <subsystem_name>_type1[_type2…]_attribute?
like undertow_request_count is too broad
you mix together listeners and servlets
nonsensical combinations

Jeff Mesnil [10:38 AM]
using your suggestion makes the metric names more readable and still unique
but you can not aggregate them

Brian Stansberry [10:38 AM]
well, not unique
because type1, type2 are not unique
i.e. they are *keys* in the address parts below subsystem=undertow
undertow_servlet_request_count

Jeff Mesnil [10:39 AM]
ah sorry, I was thinking about values

Brian Stansberry [10:39 AM]
vs undertow_listener_request_count
I have no idea if that’s truly a useful pattern though

Jeff Mesnil [10:41 AM]
that’s not something that we can do in the context of MP Metrics as it requires the metric names to be unique

Brian Stansberry [10:41 AM]
@jmesnil yeah I’m thinking in terms of “if we could do what we want”

Jeff Mesnil [10:42 AM]
in that case, I would use <subsystem>_<type*>_<attribute_name> with as much type as require to disambiguate
so, undertow_unique_metric
undertow_listener_request_count
undertow_servlet_request_count

Brian Stansberry [10:43 AM]
yeah
the question then is “as much type as require to disambiguate” — how is that determined?
i guess it can just be part of the registration algorithm since it has the entire universe available to reason from

Jeff Mesnil [10:44 AM]
exactly, I’m 1st collecting the WildFly metrics (with their whole address) and then I register them
so I can check for “uniqueness” across a subtree quite easily

Brian Stansberry [10:44 AM]
cool
do you think “deployment” should be out of the picture, just labels?
i.e. start with the /subsystem=x child

Jeff Mesnil [10:46 AM]
you’re right, that’s already the case when you suggested undertow_servlet_request_count

Brian Stansberry [10:46 AM]
where that breaks down in theory is if w/o those path elements there’s not enough to create uniqueness
i guess if that happened though, the algorithm could adapt
and add another ‘type’
actually that could be the wrong thing to do as well
e.g. a ds.xml kind of case, where a DS is part of a deployment
its metrics are the same as the metrics for a DS that’s under the subsystem tree
i.e. they shouldn’t have different names, just different labels
the opposite case, where they should have different names seems unlikely
poor subsystem design

Jeff Mesnil [10:51 AM]
I think so, we have same resources across subsystems and deployments that should be treated as the same for metrics (data source, JMS destinations, etc.)

Brian Stansberry [10:51 AM]
yes
@jmesnil do you think this sounds like a good scheme to pursue; e.g. see how well it applies to the jmx-exporter config’s Rob Cernich has worked up?
@jmesnil would it be hard to work up a prototype “uniqueness” algorithm and have it dump the names to the server.log?
i.e. run that against standalone-full-ha and we get a sense of all the names
that can then be useful in reviewing Rob’s names

Jeff Mesnil [10:57 AM]
no, that something that I can do before the end of the week
Brian Stansberry [10:57 AM]
awesome :slightly_smiling_face:
@jmesnil i’m going to paste this as a comment in Rob’s PR as an FYI; we can discuss with him as well

jmesnil:
@rcernich @bstansberry  I toyed this morning with WildFly to come up with unique names for the metrics by traversing the whole resources (using the full-ha profile).

I used the pattern `<subsystem_name>_<type1>_<type2>_attribute_name` (with some caveats I describe below)

The whole list is at https://gist.github.com/jmesnil/3919eb05fe21d68aaebaa7c54a1c9a13

The file pattern is:

```
metric_name
   * <resource address of the metric>
```
Note that these are __not__ the names that the metrics subsystem is currently exposing and I'm not sure we will be able to go towards them due to Microprofile Metrics specification constraints.

# Examples:

There are 4 `request-count` metrics related to `undertow`:
* `undertow_server_ajp_listener_request_count`
* `undertow_server_http_listener_request_count`
* `undertow_server_https_listener_request_count`
* `undertow_servlet_request_count`

My pattern does not try to reduce the number of types to have unique names and I am not sure it it worth it: `undertow_server_http_listener_request_count` seems an appropriate name (the shorter name would have been `undertow_http_listener_request_count`)

For deployment/deployment resources, I strip `/deployment=*(/subdeployment=*)` part of the name, `undertow_servlet_request_count` is actually coming from the deployment resource `/deployment/*/subsystem/undertow/servlet/*`

For ejb3 `invocations`, I have these metrics:

* `ejb3_message_driven_bean_invocations`
* `ejb3_singleton_bean_invocations`
* `ejb3_stateful_session_bean_invocations`
* `ejb3_stateless_session_bean_invocations`

They read like good metric names.

# Equivalent resources in subsystem and deployment

A few resources where having the same address pattern from the subsystem tree and the deployment (eg the jms-queue metrics from [/subsystem=messaging-activemq/server/jms-queue](http://wildscribe.github.io/WildFly/14.0/subsystem/messaging-activemq/server/jms-queue/index.html), [/deployment/ subsystem=messaging-activemq/server/jms-queue](http://wildscribe.github.io/WildFly/14.0/deployment/subsystem/messaging-activemq/server/jms-queue/index.html), and [/deployment/subdeployment/subsystem=messaging-activemq/server/jms-queue](http://wildscribe.github.io/WildFly/14.0/deployment/subdeployment/subsystem/messaging-activemq/server/jms-queue/index.html)

I made the assumptions that these resources are equivalent and will expose the same metrics.
So these 3 different `jms-queue` resource all expose the same metrics (e.g. `messaging_activemq_server_jms_queue_message_count`)

(my file is not showing the 3 resource addresses from the same metric but they were all 3 taken into account)

# JGroups resource

I had to tweak my pattern to also specify the value of the rightmost address to distinguish between
Groups resources (from [/subsystem=jgroups/channel=ee](http://wildscribe.github.io/WildFly/14.0/subsystem/jgroups/channel/ee/index.html))

There is no definition of a "protocol" but only instances of it. So I had to add the name of the protocol to distinguish them:
* ` jgroups_channel_protocol_FD_ALL_id`
* `jgroups_channel_protocol_FD_SOCK_id`

From WildFly perspective, these are different resources (and they expose different metrics) so I distinguished them too.
Having said that, many of them are not actual metrics and there is some clean up to do...

# JCA resource

While writing this comment, I realize my algorithm is buggy as the rightmost element is not enough to distinguish as shown by JCA resources at http://wildscribe.github.io/WildFly/14.0/subsystem/jca/workmanager/default/short-running-threads/index.html.
The MRR has registered resource for both `/subsystem=jca/workmanager=default/short-running-threads/` and `/subsystem=jca/workmanager=*/short-running-threads/`

rcernich:
I'm not quite sure how to parse this.  I think we should take a step back and examine how the metrics might be used, e.g. maybe write some PromQL to see how they might work.  With that in mind...

'undertow_server_`

To get a total request count with your proposal:
`undertow_server_ajp_listener_request_count + undertow_server_http_listener_request_count + undertow_server_https_listener_request_count`

Using appropriate labels:
aggregated: `undertow_server_request_count`, disaggregated: `undertow_server_request_count{listener="ajp"}`, subset: `undertow_server_request_count{listener=~"http|https"}

Similarly for bean invocations:
`ejb3_invocations` - invocations across all bean types
`ejb3_invocations{message-driven-bean=~".*"}` - invocations across all message driven beans
`ejb3_invocations{statelsess-session-bean=~".*",deployment="my.war"} - invocations across all stateless sessions beans in my.war

Given the above, it seems like using a specific `type` label for beans would be easier to use than`type=name`, but...

Looking at the [Prometheus Best Practices - Metric and Label Naming](https://prometheus.io/docs/practices/naming/#metric-names), I think we are losing focus on the following regarding metric naming:

> ...should represent the same logical thing-being-measured across all label dimensions. 

Applying this, `undertow_server_request_count_total` would be the appropriate metric name, with labels used to represent the various dimensions.  The same could be said for bean invocations: the same thing is measured across all label dimensions, with those dimensions being: deployment and bean type.

rcernich:
I also got some feedback from the metrics team suggesting that having labels like `subsystem=undertow` and metric names `undertow_request_count` was redundant and unnecessary.

bstansberry:
If subsystem=undertow and undertow_request_count are redundant and unnecessary, I think that means drop the label. I don't think "request_count" truly represents the same logical thing when applied across the server. Opinions may vary, and if they do, it's good to sort that out, since if it really should be "request_count" then all the metric names become just the attribute names and all the rest is just labels.

bstansberry:
Re "undertow_server_request_count_total" I don't understand the 'total'.  The best practices doc says "Note that an accumulating count has total as a suffix, in addition to the unit if applicable."  But it does not explain what an accumulating count is nor why 'total' should be added.

This kind of detail is significant because the subsystem is going to be generating metric names from an arbitrary set of preexisting metrics, so any kind of text that is not coming from the metric definitions themselves is problematic.

rcernich:
Regarding things like `undertow_server_request_count`, the important thing is: is it the same logical thing being measured across all dimensions.  So, if request count is not logically the same for ajp, http and https listeners, then those should each be separate metrics.

Regarding the `total` suffix, you should be able to easily generate this, where necessary.  This would be applied to all COUNTER types.  If you're looking to generate all of these, you'll also have to look at adding units to the metric names.  (Note, this only applies to the Prometheus metric names, so you can use whatever you want internally, e.g. not have things like _total and _seconds, but add those in when mapping to Prometheus.  Also keep in mind that Prometheus prefers single whole units, e.g. seconds and bytes, to things like milliseconds and megabytes.)

TBH, I'm not concerned about whether or not these are generated en mass or hand coded.  It would have been easy to generate metrics from simple patterns, but this breaks down when you need to consider the metric type (e.g. gauge vs counter) and units (e.g. simple quantity vs time).

bstansberry:
For the subsystem, hand coding isn't really an option. That is, encapsulating a bunch of custom rules inside the subsystem itself is just too architecturally terrible. Exposing the mapping stuff as part of the subsystem config (i.e. some analogue to the jmx-exporter-config.yaml stuff) isn't architecturally painful but it's a ton of config that the user then has to maintain when new metrics become available.  So being able to generate based on a reasonable number of fixed rules, plus the data available to the subsystem from the EAP management model, is pretty important for the subsystem.

rcernich:
@bstansberry, agreed.  I think the bigger question is whether or not you will want to export all of these to Prometheus.  I'm not sure how much design went into the Hawkular config, which is what I based my work on, but it's clear that the weren't exporting everything.  Furthermore, there is a performance cost to doing this.  For example, doing a complete export takes 2-3 seconds, whereas what is currently configured in my config files takes ~500ms for EAP 7.1 and ~1500ms for EAP 6.4 (the 2-3s number was for EAP 7.1).  I would guess pulling directly through dmr/mp-metrics interface would be faster than through the MBeans.  Perhaps it might be good to think about what a Prometheus exporter config might look like.

rcernich:
Given my comment above about jms (queues vs topics), I'd apply the same logic to the undertow listeners.  If those are presenting different sets of metrics, it probably makes sense for them to be different metrics, as opposed to different dimensions on the same metric.  (Basically, I think we need to be intelligent about what we do and specific in how we do it, hence my earlier comment about hand coding some of this stuff.)

bstansberry:
I asked Jeff for his thoughts on this, as he's stuck with the subsystem. ;) My biggest concern is any label that is not a key/value pair from the resource address. I suspect that will prove to be a big pain to deal with and I doubt the benefit will be worth the cost.

Some element of "hand coding" may be inevitable to produce a good result, but I think that needs probably to be done in such a way that the definition of the metric within the WF management model includes the necessary data (vs external config or the subsystem itself keeping it internally). And also the need for that should be minimized as much as possible, e.g. to cases where omitting parts of the address from the name produces the best result.

In general other than the type|servlet: $1 plus name: $2 type label stuff I think this looks nice.

Re: undertow listeners, I think the different listeners are providing conceptually the same thing, so the metrics having the same name is better. Within the EAP management model, the various listener resource types all have different definitions because they have different sets of attributes, but they are all similar.

Re: the configuration of the subsystem, it's pretty simple:

```
<subsystem xmlns="urn:wildfly:microprofile-metrics-smallrye:1.0" security-enabled="false" exposed-subsystems="*"/>
```

So you can elect a set of subsystems you want, or all. (Note that subsystems themselves have flags that turn on/off recording of metrics, since that has a cost.)

rcernich:
One thing I should mention, wrt this specific PR, is that the jmx exporter is limited in the types of metrics it can support, as each attribute is processed individually.  This makes it impossible to define a metric that might be composed of multiple values (e.g. using another attribute on an MBean in a label).

I'm glad this little PR got you thinking about broader implications and how the new mp-metrics might interact with Prometheus, but given the limited scope (parity with legacy Hawkular config), do you think this is okay to be merged?

rcernich:
@bstansberry, it just struck me...would it be worth investigating using Prometheus Collector implementations for collecting metrics within the subsystems and exposing the collected metrics?  I'm assuming you have a some sort of standard interfaces for defining metrics (e.g. name, type, units, etc.).  Anyway, I'm way off topic now...

bstansberry:
@rcernich This looks great to me, except the issue of using custom labels instead of sticking to ones that represent key/value pairs in the resource address.

I have lots of experience with getting Critical bug reports from QE because something in a new thing doesn't conform to how something in an older similar thing was done. In this case the "old thing" is this PR, which will shortly end up in all EAP image streams. The new thing is the subsystem and it just isn't practical there to use custom label patterns. Other aspects of the metric naming here (all of which I like), either we can get the subsystem to match with reasonable effort, or we'll just fight back on any bug report.

rcernich:
Understood.  The rationale behind using things like `type` is that it keeps the label set for each metric consistent.  This was based on feedback I got from the Prometheus guys.  I also suspect that it will speed things up when querying a specific type, for example:

> `ejb3_invocations{type="stateless-session-bean"}`

vs

> `ejb3_invocations{stateless-session-bean=~".+"}`

The former tests equality, while the latter is a regex.

I thought the former works/looks better from a query perspective, as it's declarative, while the latter is more like, "give me all the metrics that have a value for some label."  I'm happy to change it back if you want me to.

bstansberry:
I changed my mind based on our chat; the type: $1 stuff is fine for now while we sort of the implications of saying that a metric should have the same set of label keys for all instances.

jboss/container/eap/prometheus/config/7.1/artifacts/opt/jboss/container/prometheus/etc/jmx-exporter-config.yaml
```diff
@@ -0,0 +1,363 @@
...
+  help: Messaging JMS Queue Metrics $3
+  value: $4
+  valueFactor: 1
+  labels:
+    server: $1
+    name: $2
+  attrNameSnakeCase: true
+  type: GAUGE
+- pattern: "jboss.as<subsystem=messaging-activemq, *server=(.+), *jms-queue=(.+)><>(messages_added): (.+)"
+  name: messaging_activemq_jms_queue_$3_total
```
jmesnil: for consistency, the name should be `messaging_activemq_$3_total`

and the labels should have:

```
   type: jms-queue
```

(same remark goes for jms-topic resources)

rcernich: I wasn't sure about that, since the metric set differs between queues and topics.  I can update the PR.

PR #335, opened 2019-04-10
[CLOUD-2868] Preparations for OpenJDK 11 / RHEL8 / UBI image

jmtd:
https://issues.jboss.org/browse/CLOUD-2868


jmtd:
For existing images, this should effectively be a no-op, but we need to determine whether this causes build problems for them, with either the latest released version of Cekit or with the develop branch (or both)

jboss/container/maven/35/35scl/configure.sh
```diff
@@ -0,0 +1,25 @@
...
+chown -R jboss:root $ARTIFACTS_DIR
+chmod -R ug+rwX $ARTIFACTS_DIR
+chmod ug+x ${ARTIFACTS_DIR}/opt/jboss/container/maven/35/scl-enable-maven
+
+pushd ${ARTIFACTS_DIR}
+cp -pr * /
+popd
+
+# maven pulls in jdk8, so we need to remove them if another jdk is the default
+if ! readlink /etc/alternatives/java | grep -q "java-1\.8\.0"; then
```
luck3y: Hmm, testing this, it doesn't seem to run for me, I'm still looking at why...

luck3y: Actually I think I understand now. Moving the maven install to be before the java module so the version can be specified ends up pulling in JDK 8 as the maven dep, so then when the uninstall runs it finds jdk8 in /etc/alternatives and doesn't uninstall it.

Then JDK 11 is installed on top and we end up with both.

luck3y: Moving the maven module after the install of JDK 11 makes this DTRT and remove JDK 8 after maven is installed.

PR #347, opened 2019-09-04
[OPENJDK-38] Add support for Maven 3.6

jerboaa:
Signed-off-by: Severin Gehwolf <sgehwolf@redhat.com>

https://issues.jboss.org/browse/OPENJDK-38


jerboaa:
@jmtd Do you know whether we could potentially merge this? Would I need another review? Thanks.

jmtd:
I can merge this, but because the CCT module stuff is a bit tricky i would like to verify (either myself or confirm you have tested) the happy path for existing images doesn't change (maven 3.5 using images are still OK)

We will also need some more stuff to get Maven 3.6 images working OK (not a blocker for merging this PR as it is). As things currently stand, `jboss.container.maven.default.bash` hard-codes maven 3.5. Which might be sensible, but `jboss.container.maven.s2i.bash` depends upon `jboss.container.maven.default.bash`; and `jboss.container.java.s2i.bash` depends upon that; and our images all depend on that one (in `image.yaml`). One solution might be for us to create varieties of that whole inheritance path, but hopefully there's a better solution than that.

Or maybe we could switch the default to 3.6? We should talk to the other users of cct_module (EAP etc) and see what they think about that.

P.S.: for testing I used https://github.com/jmtd/openjdk/tree/maven36 with a local foo.repo containing the necessary private RPM repositories

jboss/container/maven/36/36scl/module.yaml
```diff
@@ -0,0 +1,22 @@
...
+  value: /opt/jboss/container/maven/36/
+- name: MAVEN_VERSION
+  value: "3.6"
+
+execute:
+- script: configure.sh
+
+packages:
+  install:
+  - rh-maven36
```
luck3y: @jmtd Will rh-maven36 ever even exist? I suspect we don't / won't need 36scl. I'm not 100% sure about this though.

luck3y:
+1 for this. For merging to master, I'd say go ahead, the stable branches shouldn't get this change though IMHO, then existing images are unaffected.

luck3y:
Sorry, didn't mean to block on my review question, LGTM, but I suspect we can just drop the scl variant completely.

jboss/container/maven/36/36scl/module.yaml
jmtd: It is currently being QA'd

jmtd: The combination of RHEL7, JDK8 and Maven 3.6 is imporant for Quarkus, alas.

jmtd:
testing:

 - [ ] RHEL 7 OpenJDK 8 Maven 3.5
 - [ ] RHEL 7 OpenJDK 11 Maven 3.5
 - [ ] RHEL 8 OpenJDK 8 Maven 3.5
 - [ ] RHEL 8 OpenJDK 11 Maven 3.5

jerboaa:
fwiw, this doesn't quite work. Will need to revise.

jerboaa:
It will need something like that on top: https://github.com/jerboaa/cct_module/commit/fe512330c69903cf35370ff98c2810185671a97d

jmtd:
! good catch, sorry I missed it. Especially before I merged :)

jerboaa:
> ! good catch, sorry I missed it. Especially before I merged :) Edit: I misunderstood. yeah this is consistent with what I was saying in [#347 (comment)](https://github.com/jboss-openshift/cct_module/pull/347#issuecomment-528351803).

PR #348 should address this.

PR #348, opened 2019-09-09
Add support for selectively using 3.6 or 3.5 maven bash

jerboaa:
The default stays at 3.5 (via epoch versioning) and can get overriden
via override files.


jerboaa:
@jmtd Please have a look. Thanks!

jerboaa:
@jmtd It looks like I won't have time to do this in the immediate future. If you want to take this over, feel free to. Otherwise I'll get back to this in ~1 weeks time or so.

PR #350, opened 2019-09-16
[CLOUD-3374] Delete old modules prior to renaming

jmtd:
https://issues.jboss.org/browse/CLOUD-3374

These modules were renamed, and the old names kept around for legacy
compatibility (containing only a dependency on the new name)


jfdenise:
@luck3y @jmtd dynamic-resources references os-java-run (that has been deleted). This module must be updated.

PR #356, opened 2019-10-02
[CLOUD-3387] Don't attempt to enable maven in run

jmtd:
https://issues.jboss.org/browse/CLOUD-3387

don't try to enable Maven (Via SCL) in the s2i run script.

We might not have provided Maven via SCL in the image anyway.


jmtd:
I've redone this (since October) to check for the script prior to sourcing, to silence the warning on RHEL8-based systems, but still enable Maven in the run environment for RHEL7-based systems. I originally just deleted the whole line, since I doubt anyone is actutally relying upon maven availability in the `run` environment, and I want to try to reduce the size and complexity of all these scripts… but this is hopefully a much easier change to review now. If/once we are no longer updating RHEL7-based images we can revisit these scripts accordingly.

jerboaa (approved):
OK.

jmtd:
Thanks!

PR #361, opened 2019-11-27
CLOUD-3392 - There is no environment variable to configure the initial metaspace size

yersan:
I'm not sure how we should proceed with these cases, please correct me if I am wrong, these changes were already merged in master but they are required in the branch used to build EAP 7.2, which seems to be 0.28.1, so hence the backporting.

Backport changes from master to 0.28.1:
PR on master was: https://github.com/jboss-container-images/jboss-eap-modules/pull/138
The original issue was: https://issues.jboss.org/browse/CLOUD-3270

This PR is required to fix https://issues.jboss.org/browse/CLOUD-3392 on EAP 7.2.x


jmtd:
Thanks for this. Code looks good but you're right we need to look at the refs situation a little more closely.

jmtd:
Are you at all able to change the ref in the legacy EAP container source away from 0.28.1? If we created a "0.28-branch" or similar for example

luck3y:
@yersan @jmtd In retrospect, creating 0.28.1 was probabyl a mistake, we should have made those named 0.28.x and then just tagged .1 .2 etc along the branch.

I think we can fix that now by just makeing the 0.28.x branch from whats in 0.28.1 and applying this change, then tagging it 0.28.2 for the next release.

What do you guys think?

yersan:
> What do you guys think?

Sounds good to me, I assume the same for the other cct_module branches that are in used (0.28.x, 0.35.x, 0.37.x)

PR #363, opened 2020-01-13
Fix Prometheus tests and re-add Jolokia tests

jmtd:
I don't think the Prometheus could ever have been run:

 • CeKit does not collect/run tests outside of "tests/features"
   for module sources (https://github.com/cekit/cekit/issues/638)
 • The features are not tagged
 • There are syntax errors in the features

Address all of the above. Tag the tests for execution on the
OpenJDK containers.

Also re-add the Jolokia tests, which were moved out of this repo and into the OpenJDK one, but should be here (IMHO) because this is the code that they test.

Other images are welcome to (and encouraged) to add their tags to these tests.

### DIFF
--- a/tests/features/jolokia.feature
+++ b/tests/features/jolokia.feature
@@ -1,0 +1,10 @@
+# Tests for jboss/container/jolokia
+@openjdk
+@redhat-openjdk-18
+@openj9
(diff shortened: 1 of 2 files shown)


jmtd:
Superceded by #364 and #369

PR #364, opened 2020-01-15
Rework Prometheus into 7, 8 modules and fix tests

jmtd:
I've substantially revised this PR since @spolti approved it, please take another look. Thanks!

Only dealing with Prometheus in this PR.

 * fix the tests which have never worked
 * delete stale README.adoc - we need a different plan for module docs
 * simplify/merge the existing prometheus modules into "jboss.container.prometheus". The existing api/bash split was gaining us nothing.
 * rename/version the existing module "7", to reflect the state of affairs from RHEL7.
 * add a new module "8.2", to reflect what to do in RHEL 8.2 and onwards: RPM based.
 * In the 7 module, install the JAR in the same place as the RPM puts it in the 8 module, and update the tests and configuration paths accordingly.


jmtd:
We will need alternative modules rather than replacements, since this will only work for RHEL 8.2+ and not earlier

jmtd:
All existing prometheus consumers will need to specify an explicit version '7' now as '8.2' will be chosen preferably by cekit. It's either that or do it the other way around. but 8.2 is the future. In practise that will mean explicitly adding a dep on this module before its dependent. E.g.

```
   install:
   - name: jboss.container.openjdk.jdk
     version: "11"
+  - name: jboss.container.prometheus
+    version: "7"
   - name: jboss.container.java.s2i.bash

```

jboss/container/prometheus/8.2/module.yaml
```diff
@@ -0,0 +1,27 @@
+schema_version: 1
+
+name: jboss.container.prometheus
+version: '8.2'
+description: ^
+  Provides support for configuring Prometheus.  Basic usage is
+  JAVA_OPTS="$JAVA_OPTS $(source $JBOSS_CONTAINER_PROMETHEUS_MODULE/prometheus-opts; get_prometheus_opts)"
+
+envs:
+- name: JBOSS_CONTAINER_PROMETHEUS_MODULE
```
luck3y: This is dropped from most of the scripting above, but its something we've documented as API -- should we support this in the scripting or mark it as deprecated here or something?

jmtd: The JAR is no longer placed under that location, but the prometheus-opts file and the jmx-exporter-config.yaml are still in there. I thought the JBOSS_CONTAINER_PROMETHEUS_MODULE variable was an informational env var rather than a config knob. Could people have been using it the other way?

luck3y: Not entirely sure, the usage in prometheus-opts suggest that they could have. It is possible I suppose that someone could have needed to run a specific version of prometheus and then they could use this mechanism. But I don't know how likely this is.

jmtd: I think it's really unlikely. and if they did, I wonder where in the image they would put it - a different place (and therefore need to change that  variable) or over the top of the one that we supplied?

jmtd:
Regards the backwards compatibility concern. At the moment (prior to this PR), we have

    JBOSS_CONTAINER_PROMETHEUS_MODULE="/opt/jboss/container/prometheus"

containing
 
    jmx_prometheus_javaagent-0.3.1.redhat-00006.jar
    jmx_prometheus_javaagent.jar -> /opt/jboss/container/prometheus/jmx_prometheus_javaagent-0.3.1.redhat-00006.jar 

Which name do you feel needs to be preserved?

 * $JBOSS_CONTAINER_PROMETHEUS_MODULE/jmx_prometheus_javaagent-0.3.1.redhat-00006.jar
 * $JBOSS_CONTAINER_PROMETHEUS_MODULE/jmx_prometheus_javaagent.jar
 * both?

Do you think this compat symlink should be in the existing module (now versioned 7), in the new one (8.2), or both? From my POV the 8.2 consumers are "new" images but I suppose the same is not true for you? I'm going to be bumping at least minor and most likely major version when I switch to RHEL 8.2.

jmtd:
I've added a compat symlink for (2) here and also in #369

jboss/container/prometheus/7/backwards_compatibility.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -ueo pipefail
+
+# Legacy location
+ln -s /usr/share/java/prometheus-jmx-exporter/jmx_prometheus_javaagent.jar \
```
spolti: if the source location don't exist this this script will fail and the image build fail, right?
maybe add a check to make sure that this will not happen?

PR #365, opened 2020-01-24
enable Maven 3.6 AppStream module prior to install

jmtd:
The AppStream module part of this commit is necessary to get access
to Maven 3.6 (instead of 3.5). The "dnf module enable" part needs
to happen before the package installation. Cekit's module ordering
places the "execute" last; so we either need to have two cekit
modules: the first to enable the AppStream module, and the second
to install the package; or (as in this commit) perform the package
installation in the script, too.

This hard-codes "dnf". However, "microdnf" does not support enabling
AppStream modules, which rules out using the minimal-flavour images
without further work.

### DIFF
--- a/jboss/container/maven/36/default/configure.sh
+++ b/jboss/container/maven/36/default/configure.sh
@@ -1,6 +1,9 @@
 #!/bin/sh
 set -e
 
+dnf module enable -y maven:3.6
+dnf install -y --setopt=tsflags=nodocs maven
+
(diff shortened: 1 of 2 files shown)


jmtd:
@jerboaa 
> Having said that, should we consider adding a feature request for cekit to support "dnf module install" via the declarative yaml?

Yes I think that's a good idea; some kind of rpm-module support or awareness in Cekit might be useful for other things too.

PR #366, opened 2020-02-07
singleton-jdk module

jmtd:
Rework the "remove surplus Java packages" logic into a single cekit module "singleton-jdk" that determins the corect versions to keep based on JAVA_VERSION and JAVA_VENDOR environment variables.

Remove the similar code from the various Maven modules.

Images using the Maven modules should now depend upon this one too and ensure it is sequenced *after* the maven modules so it cleans up afterwards.


jmtd:
Hey @luck3y I think this will be useful for the OpenJ9 images.

jboss/container/java/singleton-jdk/configure.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+set -u
+set -e
+
+# Clean up any java-* packages that have been installed that do not match
+# our stated JAVA_VERSION-JAVA_VENDOR (e.g.: 11-openjdk; 1.8.0-openj9)
```
jerboaa: Perhaps there should be a check whether `${JAVA_VERSION}` and `${JAVA_VENDOR}` is empty as the rest of the logic assumes non-empty. It would remove all JDKs in that case, AFAIUI.

jmtd: If they're undefined then the script will fail due to `set -u`. Can they be defined-but-empty? How would we end up in that situation?

jmtd: Assuming it didn't bail out, the grep pattern would be "^java--" which wouldn't match any real java packages.

jerboaa: Right, but we have it negated:
```
    if ! echo "$pkg" | grep -q "^java-${JAVA_VERSION}-${JAVA_VENDOR}"; then
```

So for any package, `grep -q "^java--"` won't match, turning that to true.

luck3y:
Aside from @jerboaa's comments, I like reducing the complexity here, so +1 from me.

jboss/container/java/singleton-jdk/configure.sh
jmtd: I see what you mean! OK I'll add a pre-check.

PR #369, opened 2020-03-10
Revamp Jolokia module, add version 8

jmtd:
Similar approach to #364 . Remove old legacy module, re-introduce tests that got removed for unknown reasons, merge down the current api/bash modules and bump the version to 7; install the JAR in the same place as the RPM puts it for RHEL 8.2+ and introduce a version 8.2 module that installs via RPM.


jboss/container/jolokia/7/module.yaml
```diff
@@ -0,0 +1,66 @@
...
+- name: AB_JOLOKIA_ID
+  description: Agent ID to use (`$HOSTNAME` by default, which is the container id).
+  example: "openjdk-app-1-xqlsj"
+- name: AB_JOLOKIA_DISCOVERY_ENABLED
+  description: Enable Jolokia discovery. Defaults to **false**.
+  example: "true"
+- name: AB_JOLOKIA_OPTS
+  description: Additional options to be appended to the agent configuration. They should be given in the format `key=value,key=value,...`.
+  example: "backlog=20"
+- name: JBOSS_CONTAINER_JOLOKIA_MODULE
```
luck3y: Not to be annoying, but this looks like the same API question as prometheus. We've documented it, so I think we should be careful in before breaking it. Should we not leave the variable in above and use /usr/share/java/jolokia-jvm-agent/ here, allowing anyone who did change this for whatever reason to not be broken by this change?

jmtd: In both cases, would you be satisfied by a compatibility symlink from the old location (under e.g. `JBOSS_CONTAINER_JOLOKIA_MODULE`) to the new one?

jmtd:
Hey @luck3y @spolti , I merged the approved #364 after tag 0.40.0 was cut. I've taken the same approach to resolve the backwards compat issue in this PR as that one. Is this OK to approve? TIA!

luck3y:
Thanks Jon, sorry for the delay in getting back to you. I think after the tag this is of less concern, so +1 from me!

Issue #370, opened 2020-03-10
different approach for docs


jmtd:
I would like to look at a scheme where we do not commit auto-generated documentation into this repository, but instead populate a separate repository with auto-generated documentation and generate a website (git site perhaps) from that. We could set up  a CI to automate the process. Travis CI or whatever.

PR #371, opened 2020-03-18
Remove autogenerated README.adoc files

jmtd:
As mentioned in #370, we need a different approach for docs. These
files get stale as the modules are updated but they are not regenerated,
and they make refactoring work very difficult.

We can either come up with a new scheme or re-introduce these later on.


luck3y:
@jmtd +1 on the stale issues, and it would be nice if these were autogenerated. Was there some tooling around this at some point?

jmtd:
> @jmtd +1 on the stale issues, and it would be nice if these were autogenerated. Was there some tooling around this at some point?

Not for automatic, but I think we could potentially hook Travis CI or similar up to a small shell script to run `./generate_docs.py` on each commit to master.  That's what I'd like to try next.

PR #373, opened 2020-03-27
New modules for ubi-minimal

jmtd:
I'm using both of these modules in order to base on the ubi-minimal base image; I need to install DNF to enable the maven module; the BZ workaround is hopefully temporary.


luck3y:
@jmtd Nice!

PR #375, opened 2020-05-19
OPENJDK-84: Always clean up S2I application source

jmtd:
Do not conditionally clean it depending on whether
S2I_ENABLE_INCREMENTAL_BUILDS is set.

To resolve the application source code always remaining in /tmp after an S2I build
<https://issues.redhat.com/browse/OPENJDK-84>

At least OpenJDK and (some) JWS images are using these particular modules. I couldn't easily figure out if EAP still does — @luck3y can you confirm?

Thanks


jmtd:
@bekionel are you OK to take a look at this one?

PR #376, opened 2020-05-19
[RHPAM-2746]Set default JVM Max memory ratio to 80% in BA Operator

desmax74:
https://issues.redhat.com/browse/RHPAM-2746
@tchughesiv @spolti @jakubschwan


spolti:
I don't think we should change things here, this can be easily changed in our side just by setting the correct env on our images.
@jmtd wdyt?

jmtd:
Yes this would change it for all dependent images, including OpenJDK and others. The BA Operator needs to override the value using a file similar to java-default-options installed by a module specific to that image (creating one if it doesn't already exist)

desmax74:
@jmtd Ok, thanks I'll close this PR

PR #378, opened 2020-06-11
[CLOUD-3484] Prometheus: don't import logging.sh

jmtd:
<https://issues.redhat.com/browse/CLOUD-3484>

We were sourcing logging.sh from the module
jboss.container.util.logging.bash module but did not have a dependency
upon it. And we aren't using it anyway.


jerboaa (approved):
OK.

PR #379, opened 2020-09-07
[OPENJDK-160] test for OPENJDK-84 (remove S2I-injected source post build)

jmtd:
tests that we remove the contents of /tmp/src after an S2I build. The code we are testing was merged here in #375


jerboaa (approved):
Looks alright to me.

PR #380, opened 2020-09-15
[OPENJDK-123] pkg-update module

jmtd:
Performs a package update (yum, or dnf, or microdnf)

### DIFF
--- a/jboss/container/util/pkg-update/execute.sh
+++ b/jboss/container/util/pkg-update/execute.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -euo pipefail
+
+for mgr in yum dnf microdnf; do
+    if test -x "/usr/bin/$mgr"; then
+        mgr="/usr/bin/$mgr"
+        break
+    fi
+done
+if ! test -x "$mgr"; then
+    echo "cannot find a package manager" >&2
+    exit 1
+fi
+
+"$mgr" update -y
--- a/jboss/container/util/pkg-update/execute.sh
+++ b/jboss/container/util/pkg-update/execute.sh
@@ -1,15 +1,14 @@
 #!/bin/bash
 set -euo pipefail
 
-for mgr in yum dnf microdnf; do
-    if test -x "/usr/bin/$mgr"; then
-        mgr="/usr/bin/$mgr"
-        break
+for candidate in yum dnf microdnf; do
+    if command -v "$candidate"; then
+        mgr="$(command -v "$candidate")"
+        "$mgr" update -y
+        "$mgr" -y clean all
+        exit
     fi
 done
-if ! test -x "$mgr"; then
-    echo "cannot find a package manager" >&2
-    exit 1
-fi
 
-"$mgr" update -y
+echo "cannot find a package manager" >&2
+exit 1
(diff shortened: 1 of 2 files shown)


jboss/container/util/pkg-update/execute.sh
```diff
@@ -0,0 +1,15 @@
...
+        mgr="/usr/bin/$mgr"
+        break
+    fi
+done
+if ! test -x "$mgr"; then
+    echo "cannot find a package manager" >&2
+    exit 1
+fi
+
+"$mgr" update -y
```
jerboaa, line 15: Isn't this missing a `[..] && $mgr -y clean all` or some such?

jboss/container/util/pkg-update/execute.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -euo pipefail
+
+for mgr in yum dnf microdnf; do
+    if test -x "/usr/bin/$mgr"; then
```
jerboaa, line 5: I'd suggest to use `command -v "$mgr"` here over `test -x` with a path. This avoids hard-coding `/usr/bin`

jboss/container/util/pkg-update/execute.sh
```diff
@@ -0,0 +1,15 @@
...
+        mgr="/usr/bin/$mgr"
```
jerboaa, line 6: Perhaps `mgr=$(command -v $mgr)` here?

jboss/container/util/pkg-update/execute.sh
```diff
@@ -0,0 +1,15 @@
...
+        break
+    fi
+done
+if ! test -x "$mgr"; then
```
jerboaa, line 10: Have you considered initializing `mgr` to say, `not-exist`, and test against that here? Like `if [ "$mgr" == "not-exist"]; then ...`

jboss/container/util/pkg-update/module.yaml
```diff
@@ -0,0 +1,7 @@
+schema_version: 1
+name: jboss.container.util.pkg-update
+version: '1.0'
+description: "Perform a package manager update."
```
jerboaa, line 4: s/Perform a package manager update/Perform a system update via the package manager/

jerboaa (changes requested):
A couple of comments. Let me know what you think.

jmtd:
@jerboaa Thank you for the review! I've pushed a commit that addresses your points, please let me know what you think.

jboss/container/util/pkg-update/execute.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -euo pipefail
+
+for candidate in yum dnf microdnf; do
+    if command -v "$candidate"; then
+        mgr="$(command -v "$candidate")"
+        "$mgr" update -y
+        "$mgr" -y clean all
```
jerboaa, line 8: Should we combine these lines? The intention for this is to clean files which got cached due to the `$mgr update -y` command, right? I.e. `"$mgr" update -y && "$mgr" -y clean all`

jmtd: The whole script gets executed in one container layer so there's no need to do that

jerboaa: OK.

jerboaa (approved):

PR #385, opened 2020-10-14
[OPENJDK-100] set networkaddress.cache.negative.ttl=0

jmtd:
Set a system property so that we do not cache negative DNS responses for
any length of time. This can interact badly with short-lived DNS names.
We can expect DNS caching to take place at the OpenShift layer anyway.

https://issues.redhat.com/browse/OPENJDK-100

### DIFF
--- a/jboss/container/java/run/bash/configure.sh
+++ b/jboss/container/java/run/bash/configure.sh
@@ -16,3 +16,13 @@ popd
 mkdir -p /deployments/data \
  && chmod -R "ug+rwX" /deployments/data \
  && chown -R jboss:root /deployments/data
+
+# OPENJDK-100: turn off negative DNS caching
+if [ -w ${JAVA_HOME}/jre/lib/security/java.security ]; then
+    # JDK8 location
+    javasecurity="${JAVA_HOME}/jre/lib/security/java.security"
+else
+    # JDK11 location
+    javasecurity="${JAVA_HOME}/conf/security/java.security"
+fi
+sed -i 's/\(networkaddress.cache.negative.ttl\)=[0-9]\+$/\1=0/' "$javasecurity"
--- a/jboss/container/java/run/bash/module.yaml
+++ b/jboss/container/java/run/bash/module.yaml
@@ -56,3 +56,4 @@ modules:
   install:
   - name: jboss.container.java.jvm.bash
   - name: jboss.container.util.logging.bash
+  - name: jboss.container.openjdk.jdk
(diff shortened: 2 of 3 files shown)


jerboaa (changes requested):
Does this actually work?

Note that `networkaddress.cache.negative.ttl` is a security property, set via `java.security` file. A default config comes with that set:

```
$ grep negative /usr/lib/jvm/java-1.8.0-openjdk/jre/lib/security/java.security 
# any negative value: caching forever
# any negative value: cache forever
# any positive value: the number of seconds to cache negative lookup results
networkaddress.cache.negative.ttl=10
```

So a simple test like this fails for me:

```
$ cat TestDNSCaching.java
import sun.net.InetAddressCachePolicy;

public class TestDNSCaching {

    public static void main(String[] args) {
        if (args.length != 1) {
            throw new AssertionError("usage: java TestDNSCaching <expected>");
        }
        int expected = Integer.parseInt(args[0]);
        int negativeVal = InetAddressCachePolicy.getNegative();
        if (negativeVal != expected) {
            throw new AssertionError("Test failed! Expected: " + expected + " but got: " + negativeVal);
        }
        System.out.println("Test passed");
    }

}
$ javac TestDNSCaching.java
$ java -Dnetworkaddress.cache.negative.ttl=0 TestDNSCaching 0
Exception in thread "main" java.lang.AssertionError: Test failed! Expected: 0 but got: 10
	at TestDNSCaching.main(TestDNSCaching.java:12)
```

Something like this should work:

```
$ echo networkaddress.cache.negative.ttl=0 > security.properties.override
# override default java.security's networkaddress.cache.negative.ttl property with 0.
# keep other properties as is
$ java -Djava.security.properties=security.properties.override TestDNSCaching 0
Test passed
```

jerboaa:
I'd suggest adding a test for this feature.

jmtd:
> Does this actually work?

You've shown that no, it doesn't! This was a rush-job, I'm sorry for wasting your reviewer time.

> I'd suggest adding a test for this feature.

Very good idea. In fact we should probably try to do this for any change, unless there's good reason not to.

I'll get this fixed up.

jmtd:
OK the test part is done, pending a merge of https://github.com/jboss-openshift/openshift-examples/pull/39 and correspodning adjustments. It fails at the moment, of course, so next, to set the property properly.

jmtd:
Hi @jerboaa I think this is now ready to be reviewed again. TIA!

jerboaa (approved):
Looks good.

PR #386, opened 2020-10-14
[rhpam-3056] Business Central remains "Container Creating" if missing known_hosts file in the githooks configuration

desmax74:
https://issues.redhat.com/browse/RHPAM-3056

@spolti @tchughesiv

### DIFF
--- a/jboss/container/user/configure.sh
+++ b/jboss/container/user/configure.sh
@@ -8,3 +8,7 @@ set -e
 groupadd -r jboss -g 185 && useradd -u 185 -r -g root -G jboss -m -d /home/jboss -s /sbin/nologin -c "JBoss user" jboss
 chmod ug+rwX /home/jboss
 chmod 664 /etc/passwd
+##we create an empty known_hosts global if the user forgot to pass the known-hosts with the id_rsa
+mkdir /etc/ssh/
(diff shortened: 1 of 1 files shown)


spolti:
we should not edit it. it should be done only on our repositories.

desmax74:
@spolti we can't add or edit under /home/jboss/.ssh because is mounted as a secret by k8s and the secret volumes are red-only (we can't change to readonly false), the other options are add under /etc/ssh or change the /etc/ssh/sshconfig to use a specific path for known_hosts, or use a option with path of known_hosts when ssh is used by the client code, do you have ideas other than these ?

spolti:
I didn't say you to do not do this, just do not use this repo to apply this change, use ours instead :)

desmax74:
@spolti sorry for the misunderstanding :) I've used this because here the user is root there isn't and the dir isn't writable by the other user, is it okay change the user there ?

spolti:
any configuration script processed by cekit will use root.

spolti:
we could, for example add this on the kieserver's configuration script and set the right permissions.

desmax74:
replaced with https://github.com/jboss-container-images/jboss-kie-modules/pull/389/

PR #391, opened 2020-12-18
Blank maven.conf file instead of removing it

jmtd:
/usr/bin/mvn complains about the missing file.

### DIFF
--- a/jboss/container/maven/8.2.3.6.8/configure.sh
+++ b/jboss/container/maven/8.2.3.6.8/configure.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+set -e
+
+# This file is shipped by a Maven package and sets JAVA_HOME to
+# an OpenJDK-specific path. This causes problems for OpenJ9 containers
+# as the path is not correct for them.  We don't need this in any of
+# the containers because ew set JAVA_HOME in the container metadata.
+# Blank the file rather than removing it, to avoid a warning message
+# from /usr/bin/mvn.
+if [ -f /etc/java/maven.conf ]; then
+  :> /etc/java/maven.conf
+fi
--- a/jboss/container/maven/8.2.3.6.8/module.yaml
+++ b/jboss/container/maven/8.2.3.6.8/module.yaml
@@ -21,3 +21,6 @@ modules:
 packages:
   install:
   - maven-openjdk8
+
+execute:
+- script: configure.sh
--- a/jboss/container/maven/8.2.3.6/configure.sh
+++ b/jboss/container/maven/8.2.3.6/configure.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+set -e
+
+# This file is shipped by a Maven package and sets JAVA_HOME to
+# an OpenJDK-specific path. This causes problems for OpenJ9 containers
+# as the path is not correct for them.  We don't need this in any of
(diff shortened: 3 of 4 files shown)


jmtd:
I've just completed a build and S2I test run for the OpenJDK images and they passed, resolving [the warning about missing files](https://github.com/jboss-container-images/openjdk/pull/172#pullrequestreview-555604393) that @jerboaa flagged.

But, I'd like it if someone on the OpenJ9 side could take a look at this too, to be sure they're happy with it, @levivic is that OK?

jboss/container/maven/8.2.3.6.8/configure.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+set -e
+
+# This file is shipped by a Maven package and sets JAVA_HOME to
+# an OpenJDK-specific path. This causes problems for OpenJ9 containers
+# as the path is not correct for them.  We don't need this in any of
+# the containers because ew set JAVA_HOME in the container metadata.
+# Blank the file rather than removing it, to avoid a warning message
+# from /usr/bin/mvn.
+if [ if /etc/java/maven.conf ]; then
```
jerboaa, line 10: This syntax is not familiar with me. Shouldn't this be `if [ ! -f /etc/java/maven.conf ]; then ...`?

jerboaa (commented):
One question:

levivic:
@mdrafiur please take a look and verify this.

jmtd:
It seems missing `/etc/java/maven.conf` is a fatal error! I missed that because it's only removed for JDK11, not 8, and I must have only tested with ubi8/openjdk-8. This revised PR blanks (not removes) the file for both JDK8 and 11.

jboss/container/maven/8.2.3.6.8/configure.sh
jmtd: Erp. Yes :)

jboss/container/maven/8.2.3.6.8/configure.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+set -e
+
+# This file is shipped by a Maven package and sets JAVA_HOME to
+# an OpenJDK-specific path. This causes problems for OpenJ9 containers
+# as the path is not correct for them.  We don't need this in any of
+# the containers because ew set JAVA_HOME in the container metadata.
+# Blank the file rather than removing it, to avoid a warning message
+# from /usr/bin/mvn.
+if [ /etc/java/maven.conf ]; then
```
jerboaa, line 10: This still doesn't look right.

jerboaa (changes requested):

jboss/container/maven/8.2.3.6.8/configure.sh
jmtd: You're right, it isn't :( I've fixed it up this morning, built and run tests over the JDK11 image .

jerboaa: OK. Looks good now.

jerboaa (approved):
Good to go.

jmtd:
@mdrafiur I've merged this but please take a look and let me know if you think this causes problems for you!

PR #395, opened 2021-03-30
[CLOUD-3951] Bump default maven version to 3.6

dbkreling:
Maven 3.5 has reached EOL and requires upgrade.

https://issues.redhat.com/browse/CLOUD-3951
Signed-off-by: Daniel Kreling <dkreling@redhat.com>


luck3y:
Note this is already fixed in master, so this change will only be applied to 0.39.x for EAP 7.3

dbkreling:
I backported commit #5eedd325d to make maven version 3.6 available, and then changed the base to 0.39.x instead of 0.39.1

luck3y (approved):

PR #406, opened 2021-06-11
ubi-clean module

jmtd:
This is a new module that removes a static list of RPMs.

The list of RPMs was determined by executing the following within the currently-shipped ubi8/openjdk-11, on a host without redhat entitlements (important): 

```
rpm -qa --queryformat "%{NAME}\n" \
    | while read pkg; do microdnf repoquery $pkg 2>/dev/null | grep -q $pkg || echo $pkg; done
```

Unfortunately we cannot run this at container build-time, because it relies upon non-UBI RPMs not being in a configured dnf repository.

### DIFF
--- a/jboss/container/util/ubi-clean/configure.sh
+++ b/jboss/container/util/ubi-clean/configure.sh
@@ -1,5 +1,12 @@
 #!/bin/sh
 
+# This list of packages was arrived at by running the following in a
+# UBI/OpenJDK container, in an environment *without any redhat entitlements* to
+# be sure the repodata is ONLY that available via normal UBI channels and not
+# the wider RHEL8 or anything else:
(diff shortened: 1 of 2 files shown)


jmtd:
I'm targetting the 0.41.x branch since that's what the shipped openjdk images are consuming.

jboss/container/util/ubi-clean/configure.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
```
jerboaa, line 2: Could you please add:
```
rpm -qa --queryformat "%{NAME}\n" \
    | while read pkg; do microdnf repoquery $pkg 2>/dev/null | grep -q $pkg || echo $pkg; done
```
as a comment here as to how this static list has been determined?

PR #423, opened 2023-09-15
[RHPAM-4777] - Dynamic resources script is reading wrong container sy…

spolti:
…s files on cgroupsv2

Please make sure your PR meets the following requirements:

- [ ] Pull Request does not include fixes for issues other than the main ticket
- [ ] Attached commits represent units of work and are properly formatted
- [ ] Every commit contains `Signed-off-by: Your Name <yourname@example.com>` - use `git commit -s`

### DIFF
--- a/jboss/container/java/jvm/bash/artifacts/opt/jboss/container/java/jvm/container-limits
+++ b/jboss/container/java/jvm/bash/artifacts/opt/jboss/container/java/jvm/container-limits
@@ -8,6 +8,19 @@
 #
 # This script is meant to be sourced.
 
+if [ "${SCRIPT_DEBUG}" = "true" ] ; then
+    set -x
+fi
+
+# query the resources based on cgroups version
+# cgroups v1 points to tmpfs
+# cgroups v2 points to cgroup2fs
+CGROUPS_VERSION="v1"
+tmp_fs=$(stat -fc %T /sys/fs/cgroup)
+if [ "${tmp_fs}" = "cgroup2fs" ]; then
+  CGROUPS_VERSION="v2"
+fi
+
 ceiling() {
   awk -vnumber="$1" -vdiv="$2" '
     function ceiling(x){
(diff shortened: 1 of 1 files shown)


jmtd:
Hey @spolti , thanks for tagging me for review. A first comment, we took a quite different approach in the openjdk images: since OpenJDK itself can now detect cgroups v1 or v2, we moved away from doing any kind of resource limit detection/setting in the wrapper scripts, and deleted container-limits script entirely. See e.g.
<https://github.com/jboss-container-images/openjdk/pull/285/files>: The JDK does it all automatically. We do change the OpenJDK default value for MaxRAMPercentage from 50% to 80%, but we do not set -Xms or -Xmx. Similarly we rely on the JVM auto-tuning for ParallelGCThreads, java.util.concurrent.ForkJoinPool.common.parallelism=, etc.

I see that *this* PR is very similar to https://github.com/asf-transfer/kogito-images/pull/1676, is that community/upstream?

spolti:
Hi @jmtd, the auto tune feature is since Java 17, right?
If so, we still on Java 11 and we do keep that container limits in case GraalVM is used as well.

Yes, the second PR is Kogito upstream.

jmtd:
> Hi @jmtd, the auto tune feature is since Java 17, right?

Cgroups V2 support is in JDK8, 11 and 17. For auto-tuning, I guess it depends which variable we are talking about, but I'm fairly sure it's been in all of them for the ones above for some time.

> If so, we still on Java 11 and we do keep that container limits in case GraalVM is used as well.

I don't really know anything about VM tuning for GraalVM so can't comment here!

spolti:
> > Hi @jmtd, the auto tune feature is since Java 17, right?
> 
> Cgroups V2 support is in JDK8, 11 and 17. For auto-tuning, I guess it depends which variable we are talking about, but I'm fairly sure it's been in all of them for the ones above for some time.

That's nice, that can be a good addition, but I am afraid that, as RHPAM is on maintenance and I am leaving the RHPAM team, anyone will be able to migrate and test it. With that said, it could be better rely on this proposed changes.
 
> I don't really know anything about VM tuning for GraalVM so can't comment here!

It is used by Kogito images (community only).

@radtriste @ricardozanini, thoughts on this? (could be a good addition for the future)



PS I don't know why I edited your comment instead replying it =/

ricardozanini (approved):

jboss/container/java/jvm/bash/artifacts/opt/jboss/container/java/jvm/container-limits
```diff
@@ -41,14 +74,27 @@ max_unbounded() {
...
+    local mem_file="/sys/fs/cgroup/memory/memory.limit_in_bytes"
+    if [ -r "${mem_file}" ]; then
+      local max_mem="$(cat ${mem_file})"
+      if [ ${max_mem} -lt ${max_mem_unbounded} ]; then
+        echo "${max_mem}"
+      fi
+    fi
+  else
+    # v2
+    local mem_file="/sys/fs/cgroup/memory.max"
```
jmtd, line 89: This path is only correct if the processes' cgroup is the root one. I.e., the result of `awk -F: '{ print $3 }' /proc/$$/cgroup`.

jmtd: It might be that this is always true in the contexts you care about, i.e., container with a java process as PID 1. It's not true on my desktop, for example, when testing the script logic:
```
$ cat /sys/fs/cgroup/memory.max
cat: /sys/fs/cgroup/memory.max: No such file or directory
$ awk -F: '{ print $3 }' /proc/$$/cgroup
/user.slice/user-1000.slice/session-300.scope
$ cat /sys/fs/cgroup/$(awk -F: '{ print $3 }' /proc/$$/cgroup)/memory.max
max
```

jmtd:
@spolti 
> That's nice, that can be a good addition, but I am afraid that, as RHPAM is on maintenance and I am leaving the RHPAM team

Good luck with your future projects! I think given you are leaving it makes a lot of sense to align this with community as closely as possible.

Since my product won't consume these changes, I haven't done an approval, but I wanted to run and test the code for you anyway, so I've done that and I think it's fine: my comment about the cgroup path is probably moot since I guess in the important container runtimes the process will be bound to the root cgroup and there won't be a heirarchy.

jboss/container/java/jvm/bash/artifacts/opt/jboss/container/java/jvm/container-limits
spolti: In containers, yes, I didn't spot any other different path, this should be enough.

luck3y (approved):

luck3y:
@spolti Are these changes not going into main as well?

spolti:
> @spolti Are these changes not going into main as well?

Hi @luck3y, I can cherry-pick, but that depends if others wants this change on main.

spolti:
@luck3y hi, are you ok if we merge this one?
And, would you like to have this cherry-picked to main?

luck3y:
lgtm @spolti, I think we should at least cherry pick to main, otherwise it'll get lost.

spolti:
> ink we should at least cherry pick to main, otherwise it'll

@luck3y https://github.com/jboss-openshift/cct_module/pull/424

PR #426, opened 2026-07-15
 BAMOE V8 - Upgrade Maven

abhijithumbe:
Pull Request: 

- Adding jboss.container.maven.39.bash.
- Updated the shared Maven module metadata to support Maven 3.9.
- Updated the Maven DNF module stream from 3.8 to 3.9.
- Updated the default Maven module dependency to use the Maven 3.9 implementation.


jboss/container/maven/default/bash/module.yaml
```diff
@@ -18,5 +18,5 @@ modules:
   install:
   - name: jboss.container.maven.api
   - name: jboss.container.java.jvm.bash
-  - name: jboss.container.maven.38.bash
+  - name: jboss.container.maven.39.bash
```
akumar074, line 21: @abhijithumbe Are we updating the default maven installation for all?

abhijithumbe: Yes. With the current change, Maven 3.9 becomes the default for all images consuming jboss.container.maven.default.bash from the new cct_module tag.

akumar074: Shouldn't this be just configurable based on name? I am assuming if other projects are using the default parameter they would need to update their config.

abhijithumbe: Do you mean jboss.container.maven.38.bash should point to maven 3.9 in the backend ?  and I think we publish a separate cct_module tag for BAMOE V8, and only the BAMOE V8 images will consume that tag. We are not changing the main branch or updating existing tag, so consumers that continue using the existing tags will not be affected.

akumar074: Okay, that makes sense. Thanks @abhijithumbe .

RishiRajAnand (approved):

RishiRajAnand:
Looks good to me

akumar074 (approved):

jfdenise (approved):

jfdenise:
@abhijithumbe would you mind squash so we can merge it. Thank-you.

abhijithumbe:
@jfdenise commits squashed. Please take a look

jfdenise:
@abhijithumbe , thank-you merged.

akumar074:
Hi @jfdenise, Can you please create a new tag `0.39.9` with this change?
cc @abhijithumbe @RishiRajAnand

jfdenise:
@abhijithumbe @akumar074 , 0.39.9 tag created.

PR #430, opened 2026-07-22
[DBACLD-240809] BAMOE V8 - Upgrade Maven (#426)

jfdenise:
Porting the fix to the 0.45.x branch, maven 3.9 becomes the default.
@abhijithumbe FYI.


jfdenise:
Actually in the context of JBOSS EAP 7.4, we are still on rhel8, so 3.8 is the target. closing this PR.

PR #431, opened 2026-07-23
Fix for JBEAP-34017, Image respin is broken, image must contain latest maven version

jfdenise:
ISSUE: https://redhat.atlassian.net/browse/JBEAP-34017


rdnovell (commented):
Looks good to me

